PR #8, opened 2020-05-15
PT-211 | Add login limit to django-admin

quyenlq:


palvelutarjotin/settings.py
```diff
@@ -218,6 +221,9 @@
     "API_KEY": env.str("LINKED_EVENTS_API_KEY"),
 }
 
+AXES_FAILURE_LIMIT = 5
+AXES_COOLOFF_TIME = 24  # hours after locked out, user will be able to attempt login
```
t0mim, line 225: Lets change this to 1 hour or to 30 minutes

PR #22, opened 2020-06-04
Fix venue query

quyenlq:
- Venue query should return empty instead of throwing error when object not found
- `p_event` should be empty when it's not created in event query
- Added `venue` to `Event` that returns venue custom data based on event `place_id`

### DIFF
--- a/graphene_linked_events/schema.py
+++ b/graphene_linked_events/schema.py
@@ -17,8 +17,12 @@ from graphene_file_upload.scalars import Upload
 from graphene_linked_events.rest_client import LinkedEventsApiClient
 from graphene_linked_events.utils import format_request, format_response, json2obj
 from graphql_jwt.decorators import staff_member_required
-from occurrences.models import PalvelutarjotinEvent
-from occurrences.schema import PalvelutarjotinEventInput, PalvelutarjotinEventNode
+from occurrences.models import PalvelutarjotinEvent, VenueCustomData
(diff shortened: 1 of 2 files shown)


graphene_linked_events/schema.py
```diff
@@ -166,9 +170,21 @@ class Event(IdObject):
...
 
     def resolve_p_event(self, info, **kwargs):
-        return PalvelutarjotinEvent.objects.get(linked_event_id=self.id)
+        try:
+            return PalvelutarjotinEvent.objects.get(linked_event_id=self.id)
+        except PalvelutarjotinEvent.DoesNotExist:
+            return None
+
+    def resolve_venue(self, info, **kwargs):
+        if self.location.id:
```
tuomas777, line 182: When using a related object's id like this, I think it is generally better to use the actual foreign key field's value, like here `self.location_id`, so you will save db hit(s).

quyenlq: Unfortunately there is no `location_id` in this `self` object because the Event obj is formatted from JSON REST response returned from LinkedEventAPI, we don't have the actual model of Event in Palvelutarjotin. `self.location.id` corresponds to  `event { location { id: ` in the json data from `LinkedEvent`

tuomas777: Aaa right, then my comment is of course invalid here, sorry I missed that :)

PR #23, opened 2020-06-04
PT-231 | Add occurrences filters by upcoming, date, time

quyenlq:


occurrences/tests/test_api.py
```diff
@@ -877,3 +877,73 @@ def test_unenrol_occurrence(snapshot, user_api_client):
...
+        start_time=datetime(2020, 1, 1, 13, 0, 0, tzinfo=timezone.now().tzinfo),
+    )
+    variables_1 = {"time": "12:00:00"}
+    variables_2 = {"time": "14:00:00+02:00"}
+    variables_3 = {"time": "11:00:00+00:00"}
+
+    executed = api_client.execute(OCCURRENCES_QUERY, variables=variables_1)
+    assert len(executed["data"]["occurrences"]["edges"]) == 1
+    executed = api_client.execute(OCCURRENCES_QUERY, variables=variables_2)
+    assert len(executed["data"]["occurrences"]["edges"]) == 1
```
tuomas777, line 919: I think most of the time with tests like this it is also good to somehow verify that the returned results are exactly the objects you are expecting, by comparing IDs or something. I've been bitten by that a few times in the past, where a test passes because the right amount of results are returned by coincidence, but actually the results have been wrong 🙂

occurrences/tests/test_api.py
```diff
@@ -686,7 +686,7 @@ def test_delete_study_group_staff_user(staff_api_client, study_group):
 
 
 def test_enrol_not_started_occurrence(snapshot, api_client):
-    # Current date is freezed on 2020-01-04:
+    # Current date is froze on 2020-01-04:
```
tuomas777, line 689: I don't normally care that match about grammar in code as long as the text is understandable (probably because I suck in it), but now that this was explicitly changed, I have to say I don't think the new version is exactly correct either 🙂 Maybe `is frozen` ?

occurrences/filters.py
```diff
@@ -0,0 +1,36 @@
...
+        fields = ["upcoming", "date", "time"]
+
+    def filter_by_upcoming(self, qs, name, value):
+        if value:
+            # Only work with PostgreSQL
+            return qs.filter(
+                **{
+                    name
+                    + "__gt": timezone.now()
+                    + timedelta(days=1) * F("p_event__enrolment_end_days")
```
tuomas777, line 29: Nice!

occurrences/tests/test_api.py
quyenlq: Damn it 🤣  It could have been either `froze` or `is frozen`

occurrences/tests/test_api.py
quyenlq: Agree! I think I missed the snapshot assertion here, there are other test cases which snapshots were used

PR #25, opened 2020-06-05
PT-270 | Add create profile mutation, support organisation modification

quyenlq:


organisations/schema.py
```diff
@@ -36,35 +36,76 @@ class Meta:
...
 
         update_object(person, kwargs)
+        if organisation_ids:
+            for org_global_id in organisation_ids:
+                org_id = get_node_id_from_global_id(org_global_id, "OrganisationNode")
+                try:
+                    organisation = Organisation.objects.get(id=org_id)
+                except Organisation.DoesNotExist as e:
+                    raise ObjectDoesNotExistError(e)
+                person.organisations.add(organisation)
```
tuomas777, line 104: Hmm, now it doesn't seem to be possible to remove organisations the person already has, just to add new ones. Is that the desired functionality?

quyenlq: No, it's not. Organisations should be cleared and re-added if necessary. Nice catch 👍

PR #31, opened 2020-06-22
PT-294 | Use APIKEY header in get query if user is staff, added showAll argument

quyenlq:
If user is staff, an authentication header will be added so user can be able to query draft events
Also added `showAll` argument to the query parameter to make draft events visible, and `publicationStatus` to filter event by `publication_status`

### DIFF
--- a/graphene_linked_events/schema.py
+++ b/graphene_linked_events/schema.py
@@ -249,6 +249,8 @@ class Query:
         text=String(),
         translation=String(),
         organisation_id=String(),
+        show_all=Boolean(),
+        publication_status=String(),
     )
     event = Field(Event, id=ID(required=True), include=List(String))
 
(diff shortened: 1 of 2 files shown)


graphene_linked_events/schema.py
```diff
@@ -291,9 +293,16 @@ class Query:
     def resolve_event(parent, info, **kwargs):
         if kwargs.get("include"):
             params = {"include": ",".join(kwargs["include"])}
-            response = api_client.retrieve("event", kwargs["id"], params=params)
+            response = api_client.retrieve(
+                "event",
+                kwargs["id"],
+                params=params,
+                auth_header=info.context.user.is_staff,
```
t0mim, line 300: Dont know about the python conventions but this ```auth_header``` could be  something like ```is_staff```. As it now tells to provide the auth header. Or we also could pass the user to client so that checks could happen in single point.

quyenlq: Good point! Updated

PR #38, opened 2020-07-03
PT-286 | Enrolment status

quyenlq:
Add status to enrolment
Update enrol occurrence mutation to support multiple occurrences enrol


occurrences/models.py
```diff
@@ -224,6 +225,15 @@ class Enrolment(models.Model):
         (NOTIFICATION_TYPE_EMAIL, _("email")),
         (NOTIFICATION_TYPE_SMS, _("sms")),
     )
+    STATUS_APPROVED = "approved"
+    STATUS_PENDING = "pending"
+    STATUS_CANCELLED = "cancelled"
+    STATUSES = (
+        (STATUS_APPROVED, _("approved")),
```
t0mim, line 232: Maybe we should have also status for declining enrolments. The use case could be when the provider declines the enrolment for some reason.

t0mim:
Maybe we should also have status for declining the enrolment. Otherwise seems good.

occurrences/models.py
quyenlq: Added `Declined` status and mutaiton to decline enrolment

PR #41, opened 2020-07-20
Update enrolment mutation

quyenlq:
Update enrolment mutation will be used from Provider UI to modify enrolment (atm end-users are not allowed to modify their own enrolment because we don't require signed-in user to enrol occurrence)

### DIFF
--- a/occurrences/tests/test_api.py
+++ b/occurrences/tests/test_api.py
@@ -1190,6 +1190,99 @@ def test_decline_enrolment(snapshot, staff_api_client, mock_get_event_data):
     assert_match_error_code(executed, API_USAGE_ERROR)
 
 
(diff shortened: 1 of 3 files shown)


occurrences/tests/test_api.py
```diff
@@ -1190,6 +1190,99 @@ def test_decline_enrolment(snapshot, staff_api_client, mock_get_event_data):
...
+      }
+      notificationType
+      status
+    }
+  }
+}
+"""
+
+
+def test_update_enrolment_unauthorized(snapshot, api_client, user_api_client):
```
tuomas777, line 1217: `snapshot` fixture doesn't seem to be needed in this test.

PR #43, opened 2020-07-28
PT-340 | Add publish event mutation

quyenlq:
- Added mutation to publish event (update `linked_event.publication_status`)
- Add `draft` argument when create event

### DIFF
--- a/common/utils.py
+++ b/common/utils.py
@@ -20,6 +20,12 @@ LanguageEnum = graphene.Enum(
     "Language", [(l[0].upper(), l[0]) for l in settings.LANGUAGES]
 )
 
+LINKED_EVENT_DATE_FORMAT = "%Y-%m-%d"
+
+
+def format_linked_event_date(datetime_obj):
+    return datetime_obj.strftime(LINKED_EVENT_DATE_FORMAT)
+
 
 def update_object(obj, data):
     if not data:
--- a/graphene_linked_events/schema.py
+++ b/graphene_linked_events/schema.py
@@ -1,7 +1,9 @@
 from django.core.exceptions import PermissionDenied
 from django.db import transaction
+from django.utils import timezone
 from graphene import (
     Boolean,
+    Enum,
     Field,
     Float,
     ID,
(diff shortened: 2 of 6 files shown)


common/utils.py
```diff
@@ -20,6 +20,12 @@
     "Language", [(l[0].upper(), l[0]) for l in settings.LANGUAGES]
 )
 
+LINKED_EVENT_DATE_FORMAT = "%Y-%m-%d"
+
+
+def format_linked_event_date(datetime):
+    return datetime.strftime(LINKED_EVENT_DATE_FORMAT)
```
tuomas777, line 27: Nitpickers would say you shouldn't have a variable named `datetime`which masks the datetime module.

graphene_linked_events/schema.py
```diff
@@ -414,10 +423,24 @@ class AddEventMutationInput(InputObjectType):
...
+    )
+
+
+class UpdateEventMutationInput(EventMutationInput):
     id = String(required=True)
 
 
+class PublishEventMutationInput(UpdateEventMutationInput):
+    start_time = String(
+        description="This will be ignore since start_time will be the current time"
```
tuomas777, line 440: Should be `ignored`. Also, I don't really understand this, if this parameter will be ignored, what is the point of having the parameter at all? 🙂

occurrences/models.py
```diff
@@ -62,6 +70,13 @@ def is_editable_by_user(self, user):
             return user.person.organisations.filter(id=self.organisation.id).exists()
         return True
 
+    def get_end_time_from_occurrences(self):
+        # Return the latest time that teacher can enrol to the event
+        if not self.occurrences.exists():
+            raise ValueError("Palvelutarjotin event has no occurrence")
+        last_occurrence = self.occurrences.order_by("-start_time").first()
```
tuomas777, line 77: `order_by("start_time").last()` would have been too simple? 😉

graphene_linked_events/schema.py
quyenlq: yeah :D It's quite confusing here because the in `UpdateEventMutationInput` that field is required. Redeclare this field there so client doesn't have to include it in the mutation

quyenlq: nvm, I think the other way around should be better to understand, fixed in the last commit

occurrences/models.py
quyenlq: Yah would have been `latest("start_time")` for maximum simplicity 😆

tuomas777: Crap, that is even better 😄 

But actually, now the whole thing would be better like
```
try:
    last_occurrence = self.occurrences.latest("start_time")
except Occurrence.DoesNotExist:
    raise ValueError("Palvelutarjotin event has no occurrence")
```
to avoid one unnecessary DB request and a theoretical race situation.

quyenlq: You are the best 🤘 Fixed!

PR #51, opened 2020-08-06
PT-345 | Add templates query to graphql api

quyenlq:
Enable `NotificationTemplate` graphql API from django-ilmoitin


common/utils.py
```diff
@@ -16,10 +14,6 @@
 )
 from palvelutarjotin.settings import REVISION
 
-LanguageEnum = graphene.Enum(
-    "Language", [(l[0].upper(), l[0]) for l in settings.LANGUAGES]
-)
-
```
a-rmz, line 22: Don't you get any schema conflict/changes by replacing the `"Language"` enum for `"NotificationTemplateLanguage"`? I get that the values are the same, but I'm not sure if you'd get any changes on the generated schema that might conflict with the front-end.

quyenlq: They are both identical, so I just deleted the one here and use the other who has been declared in `django-ilmoitin`. I don't think there is any conflict with the front-end

a-rmz: Okay 🤔 my only "concern" or thought is that it kinda couples the language enum to the `ilmoitin` package. I'm thinking that we could:
* Rename the `ilmoitin` enum to `Language` instead of `NotificationTemplateEnum`
* Add a check in `ilmoitin` that, if there's no `Language` enum, it should create it, otherwise, just use the existing one

Because I think that it's not the best idea to couple them. Maybe we could use a third opinion?

quyenlq: Alright, now I realized in 0.5.0 the `django-ilmoitin` enum is still `LanguageEnum` which is why it conflicts with the `LanguageEnum` above. It was changed in 0.5.3 to `NotificationTemplateLanguage` enum . In this case, I think it's better to keep the `palvelutarjotin` `LanguageEnum` to keep the schema consistency

a-rmz: Ahhh fair enough, good catch there. I'm still thinking it would be nice to have a single language enum to avoid duplicated values in different enums, so I'll add an issue on Ilmoitin and pick it up later. 👍

PR #58, opened 2020-08-20
PT-384 | Mass approve enrolment

quyenlq:
When approving a single enrolment, if this is a multl-occurrences event, all related enrolments of that study group to the same event should be approved as well

### DIFF
--- a/occurrences/schema.py
+++ b/occurrences/schema.py
@@ -523,7 +523,15 @@ class ApproveEnrolmentMutation(graphene.relay.ClientIDMutation):
             info, kwargs["enrolment_id"], Enrolment
         )
         custom_message = kwargs.pop("custom_message", None)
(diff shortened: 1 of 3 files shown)


tuomas777 (approved):
LGTM!

One remark, I think it is possible at least in theory to get a notification about an approved enrolment when it is actually not approved, if approving some of the later enrolments fails (and your approval is rolled back).

t0mim (approved):

quyenlq:
> LGTM!
> 
> One remark, I think it is possible at least in theory to get a notification about an approved enrolment when it is actually not approved, if approving some of the later enrolments fails (and your approval is rolled back).

Good point! It is possible to happen, though most likely it won't happen. Noted, we can come back later if the issue happens

PR #60, opened 2020-08-25
PT-393 | Add custom context template preview

quyenlq:
Added a graphql query to query template by template type (notification type). The query receive context variable as JSONString argument, then return rendered notification preview

### DIFF
--- a/common/tests/conftest.py
+++ b/common/tests/conftest.py
@@ -232,6 +232,18 @@ def notification_template_enrolment_approved_en():
         Custom message: {{ custom_message }}
         {% endif %}
 """,
+        body_html="""
+            <p>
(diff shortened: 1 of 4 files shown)


common/tests/conftest.py
```diff
@@ -232,6 +232,18 @@ def notification_template_enrolment_approved_en():
         Custom message: {{ custom_message }}
         {% endif %}
 """,
+        body_html="""
```
t0mim, line 235: One possibility would be to have a separate file for template. It might be clearer and easier to maintain.

t0mim (approved):

tuomas777 (approved):
LGTM 👍

PR #61, opened 2020-08-25
PT-385 PT-386 | Cancel occurrence mutation

quyenlq:


t0mim (approved):

t0mim (approved):

occurrences/models.py
```diff
@@ -134,6 +134,7 @@ class Occurrence(TimestampedModel):
     languages = models.ManyToManyField(
         "Language", verbose_name=_("languages"), blank=True, related_name="occurrences"
     )
+    cancelled = models.BooleanField(verbose_name=_("cancelled"), default=False)
```
tuomas777, line 137: Just an idea without knowing the context or anything: personally I prefer nullable datetimes over booleans for this kind of flags, so you'll get the timestamp of the action stored "for free".

tuomas777 (approved):
Good job! 💯

occurrences/models.py
quyenlq: Good idea. However, we don't have any use if this cancelled is a timestamp, and we also have last updated timestamp for this model. It'll be also easier for FE to just use the boolean flag to display the badge instead of the datetime value. I think we could keep it as boolean for now.

PR #67, opened 2020-09-04
PT-434 | Add seat approved, change seats taken to total seats taken

quyenlq:
Change `seat_taken` to count the total number of pending seats + approved seat
Rename the old `seats_taken` to `seats_approved`


Sartonon (approved):
Looks good!

PR #69, opened 2020-09-09
PT-443 | Add recaptcha validation

quyenlq:
Add support to verify captcha response from front-end
Specs here: https://developers.google.com/recaptcha/docs/verify


Sartonon (approved):
Looking good 👍

occurrences/schema.py
```diff
@@ -443,12 +465,20 @@ class Input:
...
+            "integration",
+        )
 
     enrolments = graphene.List(EnrolmentNode)
 
     @classmethod
     @transaction.atomic
     def mutate_and_get_payload(cls, root, info, **kwargs):
+        if settings.CAPTCHA_ENABLED:
+            verify_captcha(kwargs.pop("captcha_key", None))
```
tuomas777, line 481: Unnecessary default value.

occurrences/tests/test_api.py
```diff
@@ -1006,6 +1007,48 @@ def test_enrol_occurrence(snapshot, api_client, mock_get_event_data):
...
+                "name": "To be created group",
+                "groupSize": study_group_15.group_size,
+                "groupName": study_group_15.group_name,
+                "studyLevel": study_group_15.study_level.upper(),
+                "amountOfAdult": study_group_15.amount_of_adult,
+            },
+        }
+    }
+    executed = api_client.execute(ENROL_OCCURRENCE_MUTATION, variables=variables)
+    assert_match_error_code(executed, CAPTCHA_VALIDATION_FAILED_ERROR)
```
tuomas777, line 1046: Hmm, shouldn't this error actually be a GraphQL one saying `captchaKey` is a required field that is missing?

occurrences/schema.py
```diff
@@ -432,6 +435,25 @@ class StudyGroupInput(InputObjectType):
...
 
+def verify_captcha(key):
+    if not key:
+        raise CaptchaValidationFailedError("Missing captcha verification data")
+    secret_key = settings.RECAPTCHA_SECRET_KEY
+    verify_url = settings.RECAPTCHA_VALIDATION_URL
+
+    # captcha verification
+    data = {"response": key, "secret": secret_key}
+    resp = requests.post(verify_url, data=data)
```
tuomas777, line 446: Requests' requests (hehe) don't have a timeout at all by default, so it is [recommended](https://requests.readthedocs.io/en/master/user/quickstart/#timeouts) to always set one.

occurrences/tests/test_api.py
quyenlq: Putting the setting as part of schema parameter make it quite difficult to run during tests because Scheme is built before the settings fixture is injected in the test case. In this case, the `captchaKey` wasn't required in the schema even though that setting was set to True at the beginning of the test. I removed the `required=settings.CAPTCHA_ENABLED` and make it always optional in the mutation so it'd be easier to control the input during test.

occurrences/schema.py
quyenlq: Fixed, we have other places that also use `requests` but I'll fix them later in a separate commit

occurrences/tests/test_api.py
tuomas777: Right. I think it might have been possible to get it working using Django's [override_settings](https://docs.djangoproject.com/en/3.1/topics/testing/tools/#django.test.override_settings). Anyway, the change you made is of course fine too.

tuomas777 (approved):
Good job! 👍

PR #73, opened 2020-09-21
PT-469 | Do not allow add/update occurrences of published event

quyenlq:


occurrences/schema.py
```diff
@@ -217,9 +217,12 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
         validate_occurrence_data(kwargs)
         contact_persons = kwargs.pop("contact_persons", None)
         languages = kwargs.pop("languages", None)
-        kwargs["p_event_id"] = get_editable_obj_from_global_id(
+        p_event = get_editable_obj_from_global_id(
             info, kwargs["p_event_id"], PalvelutarjotinEvent
-        ).id
+        )
```
tuomas777, line 222: Does this work if the client does not supply argument `pEventId` ?

quyenlq: No, `pEventId` is required in both add and update mutation

tuomas777: I actually commented on the wrong part, I meant to comment about the update mutation 🙂  

So what about the update mutation, `pEventId` does not seem to be a required field there according to this line https://github.com/City-of-Helsinki/palvelutarjotin/blob/0326cd9cea7ba289c8544187b5aacdf14a554bdd/occurrences/schema.py#L250  or am I missing something?

quyenlq: Oh yeah, you are right, I didn't bother to look at the input declaration 🤦‍♂️. Even the previous code for the update mutation assumed that `pEventId` is required. Fixed now

tuomas777 (approved):
Yes yes :+1:

PR #74, opened 2020-09-21
Fix publication status query

quyenlq:
LinkedEvent API will return 404 for draft event if the query is from anonymous user so we would never get `publication_status` if it's a draft event.
Added an `is_staff` parameter to `get_event_data()` so when we want to get the `publication_status` we can use `is_staff` to always get the desired result.


tuomas777 (approved):
LGTM 👍

PR #84, opened 2020-10-16
PT-417 | Make min_group_size and max_group_size nullable

quyenlq:


occurrences/models.py
```diff
@@ -117,8 +117,12 @@ class Occurrence(TimestampedModel):
         blank=True,
         null=True,
     )
-    min_group_size = models.PositiveSmallIntegerField(verbose_name=_("min group size"))
-    max_group_size = models.PositiveSmallIntegerField(verbose_name=_("max group size"))
+    min_group_size = models.PositiveSmallIntegerField(
+        verbose_name=_("min group " "size"), blank=True, null=True
+    )
+    max_group_size = models.PositiveSmallIntegerField(
+        verbose_name=_("max group " "size"), blank=True, null=True
```
tuomas777, line 124: I think I've seen this kind of strings before somewhere...

occurrences/schema.py
```diff
@@ -406,8 +406,9 @@ def validate_enrolment(study_group, occurrence, new_enrolment=True):
     if occurrence.cancelled:
         raise EnrolCancelledOccurrenceError("Cannot enrol cancelled occurrence")
     if (
-        study_group.group_size > occurrence.max_group_size
-        or study_group.group_size < occurrence.min_group_size
+        occurrence.max_group_size and study_group.group_size > occurrence.max_group_size
```
tuomas777, line 409: After this change, if `max_group_size` is `0` the check will be skipped, which is probably not what you want.

quyenlq: Good point, but that'd be something we want. This validation should be run only if the field value > 0. Still made a change to improve readability

occurrences/schema.py
```diff
@@ -406,8 +406,8 @@ def validate_enrolment(study_group, occurrence, new_enrolment=True):
     if occurrence.cancelled:
         raise EnrolCancelledOccurrenceError("Cannot enrol cancelled occurrence")
     if (
-        study_group.group_size > occurrence.max_group_size
-        or study_group.group_size < occurrence.min_group_size
+        study_group.group_size > occurrence.max_group_size > 0
+        or 0 < study_group.group_size < occurrence.min_group_size
```
tuomas777, line 410: Hmm, doesn't this now throw when `max_group_size` or `min_group_size` is `None` ?

quyenlq: Ahhh, my bad, must be to hurry for the lunch, reverted

tuomas777 (approved):
ok let's go! 🙂

PR #91, opened 2020-11-05
PT-580 | Add categories and additional_criterial to event

quyenlq:
Screenshot 2020-11-05 at 9 19 56

UI has a need of knowing which keyword set a keyword belongs to in order to assign it to correct box in edit event page.
We need seperated fields in the event query to do so.
- For keywords of TARGET_GROUPS keyword set => already assigned in native linked event field `audiences`
- For keywords of CATEGORIES keyword set => assign to our `categories` field
- For keywords of ADDITIONAL_CRITERIA keyword set => assign to our `additional_criteria` field

### DIFF
--- a/graphene_linked_events/schema.py
+++ b/graphene_linked_events/schema.py
@@ -21,6 +21,7 @@ from graphene_linked_events.utils import (
     api_client,
     format_request,
     format_response,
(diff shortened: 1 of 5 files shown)


tuomas777 (approved):
LGTM! 👍

PR #100, opened 2020-12-14
PT-664 | Add Github Action config

quyenlq:
Basically copy paste then replace kukkuu by kultus 🙆‍♂️


tuomas777 (approved):
Looks good AFAICT 👍

You could also remove the Travis conf as there is probably no need for that anymore.

PR #102, opened 2020-12-21
[PT-666 😕 | PT-667] Count occurrence capacity by number of enrolments

quyenlq:
Previously we count the number of seats taken by the number of children enrol the occurrence.
Now added another type of occurrence where the number of seats taken increase by 1 for every approved enrolment, regardless of the size of enrolled children.


tuomas777 (approved):
LGTM 😈

t0mim (approved):

PR #103, opened 2021-01-05
PT-670  PT-669 |Add external events filter to events query

quyenlq:


nikomakela (approved):
All looks good!

quyenlq:
This PR is pending, we need more clarification from PO to show correct events. I'll reopen when we can continue developing this feature

PR #104, opened 2021-01-11
Feature/initialization fixes

nikomakela:
Some updates to a readme, added 2 missing migrations and changed the db service name in docker-compose configuration. Also changed psycopg2 to psycopg2-binary.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,17 +1,17 @@
-version: '3.7'
+version: "3.7"
 services:
-    postgres:
+    kultus-db:
         image: postgres:10
         restart: on-failure
         environment:
-            POSTGRES_USER: palvelutarjotin
-            POSTGRES_PASSWORD: palvelutarjotin
-            POSTGRES_DB: palvelutarjotin
+            POSTGRES_USER: kultus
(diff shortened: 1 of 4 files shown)


nikomakela:
Removed migrations, since they were added against (still) unsupported version of django-ilmoitin.

docker-compose.yml
```diff
@@ -1,16 +1,16 @@
-version: '3.7'
+version: "3.7"
 services:
-    postgres:
+    palvelutarjotin-db:
```
quyenlq, line 3: I'd prefer to use `kultus-db` here. `palvelutarjotin` naming is deprecated so better to avoid using it. Feel free to replace other `palvelutarjotin` with `kultus` among the rest of this file.

quyenlq:
Looks good! just a minor comment, approved already

docker-compose.yml
nikomakela: It will look a bit messy:
image

However, it's a step to a right direction. It's now done and tested:
1. Removed all old containers
2. docker-compose up --build

Result: all running well

quyenlq (approved):

PR #105, opened 2021-01-14
Amount of adults is added to calculation when counting taken or approved seats.

nikomakela:
Pull reguest for Feature/PT-680. Updated Occurrence model seat count functions and tests related to them. It was also needed to update test-snapshots.

PR #106, opened 2021-01-15
StudyGroup to support multiple StudyLevels

nikomakela:
PT-678 (and subtasks PT-686, PT-687 and PT-688) PT-679:
* A new StudyLevel model
** Migration (both directions)
** Admin UI
* Changes to GraphQL schema
** StudyGroup to support a list of StudyLevels (NOTE: This is a breaking change)
* Tests updated (and the updates includes big changes to snapshots)

It should be also noted that accepting this pull request with #105 (https://github.com/City-of-Helsinki/palvelutarjotin/pull/105) will lead to a merge conflict on test's snapshots!


occurrences/tests/snapshots/snap_test_api.py
```diff
@@ -38,7 +39,7 @@
             "name": "Increase thank certainly again thought summer. Beyond than trial western.",
             "occurrences": {"edges": []},
             "person": {"name": "William Brewer"},
-            "studyLevel": "GRADE_6",
```
nikomakela, line 41: Here is something wrong. Grade 6 should be list in a new snapshot if it was in previous.

occurrences/tests/snapshots/snap_test_api.py
```diff
@@ -19,7 +20,7 @@
                         "name": "Increase thank certainly again thought summer. Beyond than trial western.",
                         "occurrences": {"edges": []},
                         "person": {"name": "William Brewer"},
-                        "studyLevel": "GRADE_6",
```
nikomakela, line 22: Here is something wrong. Grade 6 should be list in a new snapshot if it was in previous.

occurrences/tests/test_api.py
```diff
@@ -686,7 +712,20 @@ def test_delete_venue_staff_user(staff_api_client, venue):
       groupSize
       groupName
       amountOfAdult
-      studyLevel
+      studyLevels {
+        edges {
+            node {
+                id
+                originalId
```
nikomakela, line 719: Is this originalId a good solution? In my opinion there is no need for GraphQL id at all here, because StudyLevel id is an assignable string.

occurrences/models.py
```diff
@@ -286,33 +286,31 @@ def __str__(self):
...
+    The Study Level is intended to be a hierarchical list of teaching degrees.
+    """
+
+    id = models.CharField(
+        max_length=255, primary_key=True
+    )  # PT-678 needs migration for 255 chars.
+    translations = TranslatedFields(
+        label=models.CharField(max_length=255, verbose_name=_("label"))
+    )  # Labels can have custom language translations.
+    level = models.PositiveIntegerField(
```
nikomakela, line 300: Should the name of level field be a "sort" instead? The original implementation of StudyLevels was a hierarchical tuple of tuples and this field is here only to create that same hierarchy.

occurrences/tests/test_api.py
quyenlq: Good point, I think we can just get rid of the graphQL global ID, then use `id` to display this originalId

occurrences/models.py
quyenlq: Hmn, I don't think the original levels have hierarchical order (although the naming seems really so). They are just a flat list of study levels. I'm not sure if we'll have any use of it though.

occurrences/schema.py
```diff
@@ -98,9 +98,50 @@ class PalvelutarjotinEventInput(InputObjectType):
...
+        exclude = ("id", "master")
+
+
+class StudyLevelTranslationsInput(InputObjectType):
+    language_code = LanguageEnum(required=True)
+
+
+class StudyLevelNode(DjangoObjectType):
+    # NOTE: Should the originalId have a better name?
+    originalId = graphene.String(source="pk")
```
quyenlq, line 115: As commented below, I think we can just use `ID` to display study level ID in plain text.

occurrences/tests/test_api.py
nikomakela: I found out that solution like this `variables={"id": to_global_id("StudyGroupNode", study_group.id)}` is used to create global id in tests. Didn't know that earlier. Should we use the global graphql id instead of database id? Or should we support them both?

occurrences/schema.py
```diff
@@ -475,7 +516,7 @@ class StudyGroupInput(InputObjectType):
     group_name = graphene.String()
     extra_needs = graphene.String()
     amount_of_adult = graphene.Int()
-    study_level = StudyLevelEnum()
+    study_levels = graphene.List(StudyLevelInput)
```
quyenlq, line 519: You might not want to require all fields of StudyLevelInput in this mutation. I think we can only need study_level id here.
So it could be something like:
`study_levels = graphene.List(String)`
And also the `def _get_study_levels()` will become a bit simpler

occurrences/schema.py
```diff
@@ -694,7 +735,7 @@ class Input:
         group_name = graphene.String()
         extra_needs = graphene.String()
         amount_of_adult = graphene.Int()
-        study_level = StudyLevelEnum()
+        study_levels = graphene.List(StudyLevelInput)
```
quyenlq, line 738: Same as [Line 519](https://github.com/City-of-Helsinki/palvelutarjotin/pull/106/files#diff-68d6b4eaea8b3fcb8aee6a0b456f07d442db6c5774a1e5ed978a684e739a2731R519)

occurrences/tests/snapshots/snap_test_api.py
nikomakela: Seems like this was just some random generated data that is not important here.

occurrences/tests/snapshots/snap_test_api.py
nikomakela: Seems like this was just some random generated data that is not important here.

occurrences/schema.py
```diff
@@ -827,6 +889,27 @@ def _get_or_create_contact_person(contact_person_data):
...
 
+def _get_study_levels(study_levels_data):
+    """
+    If an id is given for an instance in a list of study levels, 
+    get a list of study level instances.
+    NOTE: Creation not is not allowed.
+    """
+    result = []
+    for study_level_data in study_levels_data:
+        if study_level_data.get("originalId") or study_level_data.get("id"):
```
quyenlq, line 900: Can be simpler if we decide to get rid of the global id

occurrences/schema.py
```diff
@@ -98,9 +98,50 @@ class PalvelutarjotinEventInput(InputObjectType):
...
+        exclude = ("study_groups",)
+
+    @classmethod
+    def get_queryset(cls, queryset, info):
+        lang = get_language()
+        return queryset.language(lang)
+
+    @classmethod
+    def get_node(cls, info, id):
+        return super().get_node(info, id)
```
nikomakela, line 133: Codecov says it was not covered by tests, but is this method even needed?

occurrences/tests/test_api.py
nikomakela: oh, sorry @quyenlq. My view didn't show your response 👍

occurrences/models.py
nikomakela: When they are shown in UI, they should have a predefined order. It does not look good if there are grades from 1 to 10 and they are in random order. In my opinion there should be some hierarchy and if the levels are stored in db, the hierarchy should be given by db.

nikomakela: But if the hierarchy is only for UI, I would change the "level" to "sort"...

occurrences/schema.py
quyenlq: True, I think you can just remove this. I think this also presents somewhere in my old code :P

occurrences/tests/test_api.py
quyenlq: Yeah the goal of global ID comes from Relay, where encoded ID includes the Node type so Client-side can confirm its type immediately without further investigation. If we don't want this behavior, then we can just remove the `relay.Node` interface in the `StudyLevelNode`. One more note, we have a convention that only Types that implement `relay.Node` interface will have the suffix *Node, so consider to change the class name too :)

occurrences/models.py
nikomakela: Still thinking about this and changed my mind once again. "Sort" is a bad name because it's used by many query languages and ORMs. A "srt" or something would be better, but I think I just leave it "level", because we are really talking about study levels - not just any sortable instances, but levels of studying.

occurrences/tests/snapshots/snap_test_api.py
```diff
@@ -971,3 +1023,138 @@
...
+        }
+    }
+}
+
+snapshots["test_study_level_query 1"] = {
+    "errors": [
+        {
+            "extensions": {"code": "GENERAL_ERROR"},
+            "locations": [{"column": 13, "line": 4}],
+            "message": 'Cannot query field "edges" on type "StudyLevelNode".',
```
quyenlq, line 1157: Oops 👀

occurrences/tests/test_api.py
```diff
@@ -38,6 +44,38 @@ def autouse_db(db):
...
+            label
+            level
+            translations {
+                languageCode
+                label
+            }
+        }
+    }
+"""
+
```
nikomakela, line 78: These needs to be updated in this frontend pull request https://github.com/City-of-Helsinki/palvelutarjotin-ui/pull/66

nikomakela:
Rebased with develop.

nikomakela:
Frontend implementations
Teachers UI: https://github.com/City-of-Helsinki/palvelutarjotin-ui/pull/67
Providers UI: https://github.com/City-of-Helsinki/palvelutarjotin-admin/pull/139

quyenlq (approved):

PR #109, opened 2021-01-20
Added a new VerificationToken model which an Enrolment is using in a cancellation process.

nikomakela:
PT-712. Added VerificationToken, VerificationTokenManager and a generic
relation to Occurrences.Enrolment.


occurrences/models.py
```diff
@@ -454,6 +456,9 @@ class Enrolment(models.Model):
         verbose_name=_("status"),
         max_length=255,
     )
+    verification_tokens = GenericRelation(
+        VerificationToken, related_query_name="enrolment"
```
nikomakela, line 460: Enrolment has a field called verification_tokens, which can include all kind of verification tokens (currently there is only 1 available, which is CANCELLATION).

occurrences/models.py
```diff
@@ -454,6 +456,9 @@ class Enrolment(models.Model):
...
+    )
```
nikomakela, line 461: Enrolment has a field called verification_tokens, which can include all kind of verification tokens (currently there is only 1 available, which is CANCELLATION).

occurrences/models.py
```diff
@@ -506,3 +511,25 @@ def decline(self, custom_message=None):
...
             enrolment=self,
         )
+
+    def create_cancellation_token(self):
+        """
+        Create a cancellation verification token for an enrolment.
+        """
+        return VerificationToken.objects.create_token(
+            self, self.person.user, VerificationToken.VERIFICATION_TYPE_CANCELLATION
+        )
```
nikomakela, line 521: Cancellation tokens can be created via an enrolment instance.

occurrences/models.py
```diff
@@ -506,3 +511,25 @@ def decline(self, custom_message=None):
...
+        URL needs an cancellation token, so one will be created
+        and any active cancellation tokens related to enrolment
+        will be set in_active.
+        """
+        return VerificationToken.objects.deactivate_and_create_token(
+            self,
+            self.person.user,
+            verification_type=VerificationToken.VERIFICATION_TYPE_CANCELLATION,
+            deactivate_all_types_of_token=False,
+        )
```
nikomakela, line 535: Cancellation urls can be created via enrolment instance. It will always create a new cancellation token and deactivate all the old active cancellation tokens.

occurrences/tests/test_models.py
```diff
@@ -44,6 +44,17 @@ def test_enrolment_creation(mock_get_event_data):
...
+def test_enrolment_create_cancellation_token():
+    enrolment = EnrolmentFactory()
+    user = enrolment.person.user
+    token = enrolment.create_cancellation_token()
+    assert token.user == user
+    assert token.key is not None
+    assert token.content_object.__class__ == Enrolment
+    assert token.content_object.id is not None
+
+
```
nikomakela, line 57: Test for a cancellation token creation.

palvelutarjotin/settings.py
```diff
@@ -175,6 +175,7 @@
     "utils",
     "organisations",
     "occurrences",
+    "verification_token",
```
nikomakela, line 178: I first started implementing tokens to occurrences, but then I felt that it's almost the same to create them more generic. Now different kind of verification tokens can be created for any kind of model instances.

verification_token/models.py
```diff
@@ -0,0 +1,132 @@
...
+class VerificationTokenManager(models.Manager):
+    def deactivate_token(self, obj_or_class, key=None, verification_type=None):
+        """
+        Deactivate a token. Key parameters is used and must be given
+        when using a model class instead of model instance.
+        """
+        self.filter_active_tokens(obj_or_class, key, verification_type).update(
+            is_active=False
+        )
+
```
nikomakela, line 23: 2 different ways to use this: with an actual model instance or with a class and a key. If an object instance is used, the key-argument is needless and does nothing. If a verification_type is given, only tokens with given type are updated.

verification_token/models.py
```diff
@@ -0,0 +1,132 @@
...
+    ):
+        """
+        Deactivate old tokens (of a type) and create a new one.
+        """
+        if deactivate_all_types_of_token:
+            self.deactivate_token(obj, verification_type=None)
+        else:
+            self.deactivate_token(obj, verification_type=verification_type)
+
+        return self.create_token(obj, user, verification_type, expiry_days)
```
nikomakela, line 40: Deactivates all old tokens of a given type related to an model instance and also creates a new one. If using deactivate_all_types_of_token -argument, all type of tokens can be deactivated at once.

verification_token/models.py
```diff
@@ -0,0 +1,132 @@
...
+
+    objects = VerificationTokenManager()
+
+    def __unicode__(self):
+        return self.key
+
+    @classmethod
+    def generate_key(cls):
+        """ Generates a new key for a verification token. """
+        return token_urlsafe(getattr(settings, "VERIFICATION_TOKEN_LENGTH", 32))
```
quyenlq, line 124: Would it be too long for the default length to be 32? I imagine the verification code is something short (4-6 digits) which user can type it manually

occurrences/models.py
nikomakela: now it returns the token, but later it will return the actual link!

verification_token/models.py
quyenlq, line 40: Would it work if `verfication_type` is `None` ?

verification_token/models.py
```diff
@@ -0,0 +1,132 @@
...
+        obj,
+        user,
+        verification_type,
+        expiry_days=None,
+        deactivate_all_types_of_token=False,
+    ):
+        """
+        Deactivate old tokens (of a type) and create a new one.
+        """
+        if deactivate_all_types_of_token:
```
quyenlq, line 35: Do you think if we just check if `verifcation_type is None` is enough? So if `verficattion_type==None`, then it already meant `deactivate_all_types_of_token==True`.
Would it be better if we get rid of `deactivate_all_types_of_token` and set `verfication_type` default to be `None`?

verification_token/models.py
```diff
@@ -0,0 +1,132 @@
...
+            self.deactivate_token(obj, verification_type=verification_type)
+
+        return self.create_token(obj, user, verification_type, expiry_days)
+
+    def create_token(self, obj, user, verification_type, email=None, expiry_days=None):
+
+        key = self.model.generate_key()
+
+        if expiry_days is None:  # can be False
+            expiry_days = getattr(settings, "VERIFICATION_TOKEN_VALID_DAYS", 14)
```
quyenlq, line 47: I think we can just move this to the default argument, something like: 
`def create_token(self, obj, user, verification_type, email=None, expiry_days=getattr(settings, "VERIFICATION_TOKEN_VALID_DAYS", 14)` a bit ugly though :P ignore this if you disagree

quyenlq: Also I think the code should have a very short lifetime (15 mins)

verification_token/models.py
```diff
@@ -0,0 +1,132 @@
...
+
+class VerificationToken(models.Model):
+
+    VERIFICATION_TYPE_CANCELLATION = "CANCELLATION"
+
+    VERIFICATION_TOKEN_TYPE_CHOICES = [
+        (VERIFICATION_TYPE_CANCELLATION, _("Cancellation")),
+    ]
+
+    created_at = models.DateTimeField(verbose_name=_("created at"), auto_now_add=True)
```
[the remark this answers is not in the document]
nikomakela: I noticed that but I thought that we would have no need for updated_at.

palvelutarjotin/consts.py
```diff
@@ -20,3 +20,4 @@
 CAPTCHA_VALIDATION_FAILED_ERROR = "CAPTCHA_VALIDATION_FAILED_ERROR"
 UPLOAD_IMAGE_SIZE_EXCEEDED_ERROR = "UPLOAD_IMAGE_SIZE_EXCEEDED_ERROR"
 MISSING_MANDATORY_INFORMATION_ERROR = "MISSING_MANDATORY_INFORMATION_ERROR"
+INVALID_TOKEN_ERROR = "INVALID_TOKEN_ERROR"
```
nikomakela, line 23: Good addition

occurrences/schema.py
```diff
@@ -757,6 +759,56 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
...
+        token = kwargs.get("token")
+        try:
+            enrolment = Enrolment.objects.get_by_unique_id(unique_id)
+        except Enrolment.DoesNotExist as e:
+            raise ObjectDoesNotExistError(e)
+
+        if enrolment.status == enrolment.STATUS_CANCELLED:
+            raise ApiUsageError(
+                f"Enrolment status is already set to {enrolment.status}"
+            )
```
nikomakela, line 789: Nice catch!

occurrences/schema.py
```diff
@@ -757,6 +759,56 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
...
+        return enrolment
+
+
+def _verify_enrolment_token(enrolment, token):
+    try:
+        token_obj = VerificationToken.objects.get(key=token)
+    except VerificationToken.DoesNotExist as e:
+        raise ObjectDoesNotExistError(e)
+    if token_obj.content_object != enrolment or not token_obj.is_valid():
+        raise InvalidTokenError("Token is invalid or expired")
```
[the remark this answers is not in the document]
quyenlq: I think it's better to leave it here. We might want to use `enrolment.cancel()` in Provider UI later and this function should work without a valid token

occurrences/schema.py
```diff
@@ -757,6 +759,56 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
...
+        if enrolment.status == enrolment.STATUS_CANCELLED:
+            raise ApiUsageError(
+                f"Enrolment status is already set to {enrolment.status}"
+            )
+        if not token:
+            # Start cancellation process, sending email including token
+            if not enrolment.get_active_verification_tokens(
+                verification_type=VerificationToken.VERIFICATION_TYPE_CANCELLATION
+            ).exists():
+                enrolment.create_cancellation_token()
```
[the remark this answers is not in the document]
quyenlq: Aha, good catch. Fixed it

quyenlq (approved):
LGTM now! 💯

PR #110, opened 2021-01-20
Add link to cancel enrolment

quyenlq:


common/tests/notification_template_fixtures.py
```diff
@@ -13,6 +13,10 @@
     Occurrence: {{ occurrence.local_start_time.strftime('%d.%m.%Y %H.%M') }}
     {% endif%}
     Person: {{ study_group.person.email_address}}
+    {% if occurrence.p_event.needed_occurrences == 1 %}
+    Click this link to cancel the enrolment:
+    {{ enrolment.get_link_to_cancel_ui()}}
```
nikomakela, line 18: In the pull request #109 this function is named `enrolment.create_cancellation_url()`. That returns a string that starts with configuration variable `settings.VERIFICATION_TOKEN_URL_MAPPING.get("appname.model").get("CANCELLATION")` and is appended with `cancellation_token.key`.

So the full configuration is 
```
VERIFICATION_TOKEN_URL_MAPPING = {
    "occurrences.enrolment": {
        "CANCELLATION": f"{KULTUS_TEACHER_UI_BASE_URL}/cancel_enrolment_url/"
    }
}
```

occurrences/models.py
```diff
@@ -506,3 +513,14 @@ def decline(self, custom_message=None):
             custom_message=custom_message,
             enrolment=self,
         )
+
+    def get_link_to_cancel_ui(self, language=settings.LANGUAGE_CODE):
+        return (
+            f"{settings.KULTUS_TEACHER_UI_BASE_URL}{language}/enrolments/cancel/"
+            f"{self.get_unique_id()}"
+        )
```
nikomakela, line 521: In the pull request #109 the verification code link -function is named `enrolment.create_cancellation_url()`. That returns a string that starts with configuration variable `settings.VERIFICATION_TOKEN_URL_MAPPING.get("appname.model").get("CANCELLATION")` and is appended with `cancellation_token.key`.

So the full configuration is 
```
VERIFICATION_TOKEN_URL_MAPPING = {
    "occurrences.enrolment": {
        "CANCELLATION": f"{KULTUS_TEACHER_UI_BASE_URL}/cancel_enrolment_url/"
    }
}
```

In that solution, the path can be configured "fully" in settings, which at least helps managing. I'm not sure yet how the server instances are usually built in Helsinki dev projects, but in my experience there can be different kind of url-paths in different servers - e.g. a client, network, area or something can be added to a path as an extra. This sure looks cleaner, thats certain! Let me know what we decide so there are not multiple ways to do it.

nikomakela: the verification code (`enrolment.create_cancellation_url()`) might actually not need that mapping at all, because it could just add a parameter to a link created by `enrolment.get_link_to_cancel_ui()`. But it could be used by `get_link_to_cancel_ui`.

quyenlq: good point! In fact, the `KULTUS_TEACHER_UI_BASE_URL` is already different between staging and production, but it won't work if the structure of the URL is different. I'll update this

nikomakela (approved):

PR #113, opened 2021-01-26
VerificationToken now has a nullable person field instead of an user.

nikomakela:
PT-712. The Old implementation assumed that the person is always linked to an user instance, but it was quite the opposite, when dealing with enrolments.


nikomakela:
A screenshot from local development environment: 
image

There is no data migration, because the implementation is so new. The person field can be null, so all the old verification tokens now have a null in a person field.

quyenlq (approved):
Looks good, there is a lint check error but approved already

PR #117, opened 2021-02-01
Improved captcha handler.

nikomakela:
PT-735 UI will always send the captcha, and if it was not removed, it raised an error. Pop with None as default takes care of the issue.


quyenlq (approved):
LGTM!

PR #118, opened 2021-02-02
PT-725 | Add payment instruction to p_event

quyenlq:


nikomakela (approved):
The changes looks good to me! I'm not sure what's the difference between description and the payment instructions, but when there are 2 fields, instruction for payment can be taken care more precisely. 

In this frontend task https://helsinkisolutionoffice.atlassian.net/browse/PT-726 it was thought that we will need only one price for an event and there will be no space for other information, but maybe that changes in future.

quyenlq:
Yeah initially we wanted to include the payment instruction inside the description, but PO thinks that it'd be clearer if it stays in a separated field.

We also need a separate ticket for it to update the UI form

PR #119, opened 2021-02-03
Changes to enrolment received messages email templates

nikomakela:
PT-742 Removed the description of the message and made the titles more simple.


quyenlq (approved):
👍

PR #130, opened 2021-02-26
PT-806 | Add options to venue custom data

quyenlq:


occurrences/schema.py
```diff
@@ -391,6 +395,10 @@ class Input:
         has_clothing_storage = graphene.Boolean()
         has_snack_eating_place = graphene.Boolean()
         outdoor_activity = graphene.Boolean()
+        has_toilet_nearby = graphene.Boolean()
+        has_area_for_group_work = graphene.Boolean()
+        has_indoor_playing_area = graphene.Boolean()
+        has_outdoor_playing_area = graphene.Boolean()
```
nikomakela, line 401: The whole schema file is really long and is not DRY. However, I think this is the best option this time.

nikomakela (approved):

occurrences/schema.py
quyenlq: True, we didn't have information in the beginning that how many options will be. If the number of these boolean fields increases one more time, we need to find a better way to do it. May be combining all booleans into one integer is a good solution

PR #131, opened 2021-03-12
Occurrence signal is updating LinkedEvent Event languages.

nikomakela:
A pull request https://github.com/City-of-Helsinki/palvelutarjotin-admin/pull/156 has changes related to this pull request. With these PRs the goal is to remove language selection from provider UI's event form. The event form mutates an event in LinkedEvent. The state would now be maintained and automatically synced to LE by a signal receiver that listens changes on occurrences' languages.

PT-854  Added a signal  receiver for Occurrence which is called post_save and post_delete. It should sync an event occurrences languages to an event in LinkedEvents API.

PT-857 A migration script that adds Arabic, Russia and Chinese to Languages. Occurrence schema is changed to fetch languages from Language table instead LanguageEnum used by other implementations. Languages can now be queried also.
Fixed tests after language changes.


nikomakela:
During my tests, the languages were always synced properly to LE, but there is a minor downside in this approach. Removing and adding languages to occurrences have a different signal, so if some languages are removed and changed to new ones, 2 REST requests will be made to LE. If an language that is linked to a multiple occurrences is removed from one of them, no call to LE is made, since the state should not change. However, a call is made if a language that already exists in some occurrences is added to another. Signals just didn't give enough information so I could have avoided these.

nikomakela:
With this solution the event selection from event form can be removed, but there is a downside in using those signals in API. Another approach would be to keep the language selection in event form, but to filter languages in occurrences language selection by a selection made to an event. This is easier for API, but worse in UX sense.

occurrences/tests/test_api.py
```diff
@@ -504,7 +527,12 @@ def test_update_occurrence_of_published_event(
```
quyenlq, line 528: I think can also add tests for languages query as well

occurrences/schema.py
```diff
@@ -965,17 +989,14 @@ def _get_or_create_contact_person(contact_person_data):
...
-def _get_study_levels(study_level_ids):
-    """
-    Get a list of study level instances by using a list of study_level_ids.
-    NOTE: Creation not is not allowed.
-    """
+def _get_instance_list(ModelClass, instance_pks: List[str]):
+
     result = []
-    for study_level_id in study_level_ids:
+    for instance_pk in instance_pks:
```
quyenlq, line 995: I think `ModelClass.objects.filter(pk__in=instance_pks)` works too

occurrences/signals.py
```diff
@@ -1,10 +1,45 @@
...
+    p_event: PalvelutarjotinEvent, event_language_ids: List[str]
+) -> None:
+    """
+    Update event languages to LinkedEvents.
+    """
+
+    # FIXME: This is a needless call if LE would not need a full event data.
+    # Since the LE needs at least all the required fields when an event is updated,
+    # we first need to fetch the current event object.
+    event_obj = json.loads(
```
quyenlq, line 26: Good news, we already have 
`p_event.get_event_data()` for this

occurrences/signals.py
```diff
@@ -43,3 +78,40 @@ def remove_message_id(sender, message, **kwargs):
...
+def update_event_languages(sender, instance, action, pk_set, **kwargs):
+    """
+    When occurrence languages updates,
+    the languages should be synced to LinkedEvents Event languages.
+
+    NOTE: This will be called once by remove actions and once by additions.
+    """
+    # When languages are cleared, an empty list can be sent
+    if action == "post_clear":
+        send_event_languages_update(instance.p_event, [])
```
quyenlq, line 94: Hmn, not sure if I get it right, does this mean if single occurrence languages are cleared, event language will become empty? If so it doesn't seem right because other occurrences can still have some languages left. 
In addition, an event cannot have empty `in_language` so this will also raise error when calling LE

occurrences/signals.py
```diff
@@ -43,3 +78,40 @@ def remove_message_id(sender, message, **kwargs):
     # ESP in order to avoid the message being identified as spam by some strict
     # filters.
     message.extra_headers.pop("Message-ID", None)
+
+
+@receiver([m2m_changed], sender=Occurrence.languages.through)
```
quyenlq, line 83: Not sure if this receiver also covers the case when an occurrence is removed? What would happen then

occurrences/signals.py
```diff
@@ -43,3 +78,40 @@ def remove_message_id(sender, message, **kwargs):
...
+@transaction.atomic
+def update_event_languages(sender, instance, action, pk_set, **kwargs):
```
quyenlq, line 85: I think we should have a test for this too

occurrences/tests/test_api.py
quyenlq: because I'm obsessed with getting high test coverage 😜

occurrences/signals.py
nikomakela: I tried **HARD** using that, but I couldn't find out how I could reassign InLanguages there! The problem was that the get_event_data function creates X-objects which I found out to be readonly objects. So, I needed to call the api_client  fetcher here and restructure it differently.

occurrences/signals.py
nikomakela: Nice catch! This was the action I didn't even test yet, because I thought it would be that simple :D. I need to add tests for all the 3 actions.

occurrences/schema.py
nikomakela: It does work, but the filter does not raise an exception and that was the reason why I chose this approach. Should it raise, what do you think?

occurrences/signals.py
nikomakela: Need to check that! Well thought. I could even add an unit test for it.

occurrences/tests/test_api.py
nikomakela: Absolutely! I didn't want to write them before your first look! :) They would have gone to a trash folder if you would have rejected this. That's why I left this PS as a draft.

occurrences/signals.py
nikomakela: actually, after `_asdict()` call I could reassign the value, but I couldn't find out how to restructure it as basic json that LE requests, without breaking the data integrity.

occurrences/schema.py
quyenlq: Ah you so want to raise an exception if any id in the list doesn't exist?

nikomakela: In my opinion, if you submit a form with an invalid value, an error should be raised. The method is now used with study levels and languages, which both have a quite static list of ids in a string format. I would say it's better that an error is raised and the save process is rejected, if an user wants to select an item that is not supported. Otherwise he would miss the information that some of his inputs wasn't actually saved. So yes, I would say it's better that we raise an exception if any id in the list doesn't exist.

quyenlq: Sure, totally valid reason. I didn't think it through, please ignore my suggestion above

occurrences/models.py
```diff
@@ -120,17 +132,16 @@ def get_link_to_teacher_ui(self, language=settings.LANGUAGE_CODE):
...
-    name = models.CharField(verbose_name=_("name"), max_length=20)
-
-    def __str__(self):
-        return self.name
-
-    class Meta:
-        verbose_name = _("language")
-        verbose_name_plural = _("languages")
+    def get_event_languages_from_occurrence(self):
+        return list(
```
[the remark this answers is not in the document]
nikomakela: Changed

occurrences/signals.py
nikomakela: Couldn't improve. Chatted about this in slack!

occurrences/signals.py
nikomakela: Using event_language_ids instead, as is used always with post_add.

occurrences/signals.py
nikomakela: Added some tests.

occurrences/signals.py
nikomakela: Added a new signal receiver and a test for that one.

occurrences/factories.py
```diff
@@ -11,6 +12,14 @@
 from organisations.factories import OrganisationFactory, PersonFactory
 
 
+class LanguageFactory(factory.django.DjangoModelFactory):
+    id = factory.Faker("pystr", max_chars=10)
+    name = factory.Faker("text", max_nb_chars=20)
+
+    class Meta:
+        model = Language
+
```
quyenlq, line 21: Should we add some default sorting here so FI SV and EN will be always on top?

quyenlq (approved):
Very well done! 👌 Leave a comment about sorting languages, feel free to ignore it if you think it's unnecessary

occurrences/factories.py
nikomakela: I had made a ticket about that, but when I started to implement it, I started thinking that API should not make such a decision for UI. I made that feature to provider's UI instead and it is included in https://github.com/City-of-Helsinki/palvelutarjotin-admin/pull/156.

nikomakela:
Thanks! Let's wait for UI pull request to be approved.

nikomakela:
Because of changes in language query, we actually need to change teachers UI queries also!

PR #138, opened 2021-03-30
PT-909 | Update README

quyenlq:


nikomakela (approved):
The changes has coverage for all that was in my mind, so the changes are good and approvable. A readme and any other free text is something that could be improved forever. If new ideas or any need for updates comes up, this could always be improved with a light mind.

nikomakela:
The next step could be a bootstrap script or a working docker composer environment. This sure helps in both!

PR #139, opened 2021-04-08
Occurrence mutation should validate the start time against enrolment

nikomakela:
PT-934. As a provider when I mutate occurrence to API, it should validate the start time against enrolment

Enrolment summary enrolments are ordered by status to get same snapshot every time when running pytest.

NOTE: New design for occurrences could be waited before resolving this pull request, because it's been already talked that occurrence start and end time fields could be changing.  They should not affect on API, but who knows...

### DIFF
--- a/occurrences/schema.py
+++ b/occurrences/schema.py
@@ -202,7 +202,7 @@ class OccurrenceNode(DjangoObjectType):
         return self.amount_of_seats - self.seats_taken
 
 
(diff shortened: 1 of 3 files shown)


quyenlq (approved):
Looks good to me!

nikomakela:
After some clarifications and discussions, this should be good to be merged.

PR #141, opened 2021-04-09
PT-915 PT-916 | Change the way of counting seats taken

quyenlq:


occurrences/models.py
```diff
@@ -519,6 +517,17 @@ def send_event_notifications_to_contact_people(
             )
 
     def approve(self, custom_message=None):
+        group_size = (
```
quyenlq, line 520: Since now pending enrolment is not count to the seats taken immediately, we need to move the group size validation into `.approve` function

occurrences/schema.py
```diff
@@ -502,15 +502,6 @@ def validate_enrolment(study_group, occurrence, new_enrolment=True):
             raise EnrolmentMaxNeededOccurrenceReached(
                 "Number of enrolled occurrences is greater than the needed occurrences"
             )
-        group_size = (
```
quyenlq, line 505: This cannot be here anymore because we should allow new enrolment regardless is group size, as long as it's not approved. This validation need to be moved to inside `enrolment.approve`

nikomakela (approved):
Approved! The new wizard is better in UX wise, but in API the data integrity is checked later, which is bad for the data. However, the validation still exists and its done before the publishing which is the most important thing.

PR #142, opened 2021-04-09
PT-942 | Make location optional in event mutations

quyenlq:
Since location selector is moved to Step 2 in the new event creation wizard, it should be possible for FE to create draft event without location. GraphQL API should make this field optional in the mutation


nikomakela (commented):
Good job. Approved.

nikomakela (approved):

PR #143, opened 2021-04-14
PT-878 | Mass approvement and send message to multiple enrolments 

quyenlq:
- I also update add event mutation to only allow auto_acceptance=false if and only if needed_occurrences=1


nikomakela (commented):
More tests should be added because they also act as specifications. Me as an example, I had heard (multiple times) that manual approval is not allowed for multi-occurrence enrolments, but I really understood it only now when the validation errors where highlighted. :)

nikomakela (approved):
Looking good!

nikomakela:
This feature is needed and it looks ready, so I will merge it.

PR #144, opened 2021-04-14
PT-921 | Allow event without endtime

quyenlq:


occurrences/models.py
```diff
@@ -121,7 +121,8 @@ def get_end_time_from_occurrences(self):
         try:
             last_occurrence = self.occurrences.latest("start_time")
         except Occurrence.DoesNotExist:
-            raise ValueError("Palvelutarjotin event has no occurrence")
+            # Return None if event has no occurrences
+            return None
```
nikomakela, line 125: It's important that we have this same thing tested when publishing the event. I found the implementation about that, but I could not find any tests written for it.

```
def _prepare_published_event_data(event_id):
    # Only care about getting published event data, no permission/authorization check
    # here
    p_event = PalvelutarjotinEvent.objects.get(linked_event_id=event_id)
    if not p_event.occurrences.exists():
        raise ApiUsageError("Cannot publish event without event occurrences")
```

nikomakela (approved):
Changes are simple and approvable. More tests to event publishing could be written.

occurrences/models.py
quyenlq: Good point! This part is executed in PublishMutation tests but the output hasn't been confirmed because of the data mock-up thing. Should be a separate unit test for this. Working on it now...

nikomakela:
Before merging this PR, it should be first checked if this is really needed (after all). The team is talking about the possibility that there should always be at least 1 occurrence. If the occurrence exists, we would also always have that ending time!

quyenlq:
Closed as we don't need this feature anymore

PR #148, opened 2021-04-21
As an admin I can export organisations persons

nikomakela:
PT-960. Organisations persons can be exported from admin site. They can be rendered using a new admin template that includes a table of all the persons with a joined_date or a CSV file could be created.

image


t0mim (approved):

nikomakela:
I had some problems to create / run unit tests for the new reports app views, but I created a new task for it: PT-997.

PR #149, opened 2021-04-22
Improvements to admin filters

nikomakela:
PT-983. Improved list fields, list filters and search fields on occurrence and enrolment admin sites.

image

image

image


nikomakela:
Pomodori team is lacking another BE coder at the moment (vacation), so since these are very basic Django admin site changes and they don't include any new joined queries, I will self approve this. The changes are much needed to maintain production database.

PR #151, opened 2021-04-23
Fixed user saving on admin site when there is no link to a person instance.

nikomakela:
PT-994. Use username in user listing when name and email are not available


nikomakela:
Since this is unit tested and the changes are like basic Python and Django changes, to save valuable time now when budget is tight, I will self approve this. The bug is blocking the use and production kick-off.

PR #152, opened 2021-04-26
Changes to approved notification template.

nikomakela:
PT-916 Changed the main text.

PT-966, PT-916. If the UI is not able to implement event cancellation feature at this time, should the cancellation link be removed from notification template.


nikomakela (commented):
Review by myself

Event name:

- Finnish is Finnish
- Swedish name is tried and Finnish is used if Swedish is not available.
- English name is tried and Finnish is used if English is not available.

Times:
- Finnish has "klo" between date and time
- Swedish has "kl" between date and time
- English has nothing between date and time.

Cancellation link is removed.

PR #153, opened 2021-04-26
Changes to rejection notification templates.

nikomakela:
PT-917.


nikomakela (commented):
Review by myself 

Event name:
- Finnish is Finnish
- Swedish name is tried and Finnish is used if Swedish is not available.
- English name is tried and Finnish is used if English is not available.

Times:
- Finnish has "klo" between date and time
- Swedish has "kl" between date and time
- English has nothing between date and time.

PR #154, opened 2021-04-30
Using custom trans function to get fields translated value to notifications.

nikomakela:
PT-1002. Added a custom trans function to django-ilmoitin notification
context.

The notification templates were written so that Finnish was used as a fallback language, like this: `event.name.sv if event.name.sv is defined else event.name.fi`. If the event was created in e.g with Swedish UI, only event.name.sv was defined! Then, if the notifications were requested in Finnish or English, an error was raised. If the event was created with Finnish UI, there would not have been any problems, because event.name.fi, which was used as a fallback, was defined.

Django-ilmoitin context does not have a real Event instance so it's translation functions cannot be used. The jinja environment it uses does not include Django templatetags. Because of those reasons, it was needed to create own language translation handler, which uses Parler fallback languages configuration.


nikomakela:
I'm not sure if the translation could be always in the context of django-ilmoitin if it is inserted in a dummy_context here https://github.com/City-of-Helsinki/palvelutarjotin/blob/master/occurrences/notifications.py. Django-ilmoitin readme about it is here: https://github.com/City-of-Helsinki/django-ilmoitin#usage.

occurrences/tests/snapshots/snap_test_api.py
```diff
@@ -295,7 +295,7 @@
                         {
                             "node": {
                                 "id": "grade_1",
-                                "label": None,
+                                "label": "first grade",
```
nikomakela, line 298: I don't know why these are in this PR :D Should they be or should they not?

quyenlq (approved):
LGTM!

occurrences/tests/snapshots/snap_test_api.py
quyenlq: They should belong to the update translations of study levels PR I guess. But let's just keep it here, otherwise the test will fail

PR #161, opened 2021-05-05
Pre realease 0.3.1

quyenlq:


nikomakela (approved):
Looks good. I may have forgot to mention that I have already collected some release notes here: https://helsinkisolutionoffice.atlassian.net/browse/PT-998 😂 

I think them all are included.

nikomakela:
We could start using some automation for release notes. We could use this as standard: https://www.conventionalcommits.org/en/v1.0.0/

PR #163, opened 2021-05-10
Added enrolment id and email contact info to events' enrolments report.

nikomakela:
PT-1025. Added 2 new columns to both report formats: html and csv.


quyenlq (approved):
Looks good! 👌

PR #165, opened 2021-05-11
Events' enrolments report can be filtered with global ids.

nikomakela:
PT-1031.  Added ability to fetch events' enrolments report with global (graphene) id.
PT-997. Added unit tests for report views.

### DIFF
--- a/reports/tests/test_views.py
+++ b/reports/tests/test_views.py
@@ -14,7 +14,7 @@ from occurrences.factories import (
     StudyGroupFactory,
 )
 from occurrences.models import Enrolment, PalvelutarjotinEvent
-from organisations.factories import OrganisationFactory, PersonFactory
+from organisations.factories import OrganisationFactory, PersonFactory, UserFactory
 from organisations.models import Organisation
 from reports.views import (
     ExportReportViewMixin,
@@ -23,6 +23,8 @@ from reports.views import (
     PalvelutarjotinEventEnrolmentsAdminView,
     PalvelutarjotinEventEnrolmentsMixin,
 )
+from rest_framework import status
+from rest_framework.test import APIClient
 
 
 class ExportReportViewMixinTest(TestCase):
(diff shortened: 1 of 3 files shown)


reports/tests/test_views.py
```diff
@@ -1,7 +1,266 @@
...
+
+    @pytest.mark.django_db
+    def test_get_queryset(self):
+        (
+            p_event1,
+            p_event2,
+            p_event3,
+            p_event4,
+        ) = PalvelutarjotinEventFactory.create_batch(4)
+        self.view.request = RequestFactory().get(
```
quyenlq, line 46: I think we can also test the filter here by creating 5 p_events but only query for 4 
You can also use `snapshot` fixture to match the data instead of only relying on the number of p_events return

reports/tests/test_views.py
```diff
@@ -1,7 +1,266 @@
...
+
+        # Setup request and view.
+        self.view = self.DummyOrganisationPersonsMixinView()
+
+    @pytest.mark.django_db
+    def test_get_queryset(self):
+        org1, org2 = OrganisationFactory.create_batch(2)
+        PersonFactory.create_batch(2, organisations=[org1])
+        PersonFactory.create_batch(2, organisations=[org2])
+        self.view.request = RequestFactory().get(
```
quyenlq, line 68: Same here. It'd be better to create negative data and test if it doesn't appear in the result

reports/tests/test_views.py
```diff
@@ -1,7 +1,266 @@
-# Could not get this work yet:
-# - "axes.exceptions.AxesBackendRequestParameterRequired:
-# AxesBackend requires a request as an argument to authenticate"
```
quyenlq, line 3: The `AxesBackend` used in our API in order to prevent brute force attacks to `django-admin`.
There seems to be a [compatible issue](https://github.com/jazzband/django-axes/issues/340) with pytest `admin_client` fixture and the `AxesBackend` authentication backend.
I think the quickest solution here is to remove the `"axes.backends.AxesBackend"` from `AUTHENTICATION_BACKENDS` test setting (using `override_settings`) when running the test.

quyenlq (approved):
Looks good to me, some minor comments. Feel free to ignore if any of them is unnecessary :)

reports/tests/test_views.py
nikomakela: Ain't this quite the same without snapshots? 
```
    def test_get_queryset(self):
        (
            p_event1,
            p_event2,
            p_event3,
            p_event4,
        ) = PalvelutarjotinEventFactory.create_batch(4)
        self.view.request = RequestFactory().get(
            "/fake-path?ids=%s,%s,%s" % (p_event1.id, p_event2.id, p_event3.id)
        )
        self.assertEqual(
            set(self.view.get_queryset()), set([p_event1, p_event2, p_event3])
        )
```

reports/tests/test_views.py
nikomakela: Changed to sets comparison as in previous case.

reports/tests/test_views.py
nikomakela: Good to know! But I think I will still jump over this test at this time.

PR #166, opened 2021-05-15
PT-1035 | Add JWT Auth to Export CSV view

quyenlq:


nikomakela (approved):
We had DRF (as a dependency) already, so this is a good solution! Only a little amount of changes needed, which is good. Without DRF installed already, I wouldn't like the idea that we need to install a whole huge library of new code just to easily support JWT. I like that authentication classes can be configured per view! 👍 

Because the DRF is now chosen, I also investigated a CSV library (https://github.com/mjumbewu/django-rest-framework-csv which is preferred by DRF devs) that would really support `text/csv` -media type. Currently, if the client accepts only CSV as a response, Django responses that It's not supported. When `django-rest-framework-csv` is installed, a CSVRenderer class adds the support for Accept header. However, writing the CSV-content would need refactoring. If more CSV features are needed in future, maybe that library could be added.

PR #169, opened 2021-05-18
Occurrence cancellation notifications filtered.

nikomakela:
PT-1014. When occurrence is cancelled, enrolments that are already cancelled or declined should not receive any messages.

### DIFF
--- a/occurrences/models.py
+++ b/occurrences/models.py
@@ -241,7 +241,9 @@ class Occurrence(TimestampedModel):
     def cancel(self, reason=None):
         self.cancelled = True
         self.save()
-        for e in self.enrolments.all():
+        for e in self.enrolments.filter(
+            status__in=[Enrolment.STATUS_PENDING, Enrolment.STATUS_APPROVED]
+        ):
             e.set_status(Enrolment.STATUS_CANCELLED)
             e.send_event_notifications_to_contact_people(
                 NotificationTemplate.OCCURRENCE_CANCELLED,
--- a/occurrences/tests/test_notifications.py
+++ b/occurrences/tests/test_notifications.py
@@ -208,11 +208,20 @@ def test_cancel_occurrence_notification(
     notification_template_cancel_occurrence_en,
     notification_template_cancel_occurrence_fi,
 ):
-    study_groups = StudyGroupFactory.create_batch(3)
-    for s in study_groups:
-        EnrolmentFactory(study_group=s, occurrence=occurrence, person=s.person)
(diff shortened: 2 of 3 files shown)


occurrences/models.py
```diff
@@ -241,7 +241,9 @@ def is_editable_by_user(self, user):
     def cancel(self, reason=None):
         self.cancelled = True
         self.save()
-        for e in self.enrolments.all():
+        for e in self.enrolments.exclude(
+            Q(status=Enrolment.STATUS_CANCELLED) | Q(status=Enrolment.STATUS_DECLINED)
+        ):
```
nikomakela, line 246: Exclude cancelled and declined enrolments from occurrence cancel handler

occurrences/tests/test_notifications.py
```diff
@@ -208,11 +208,21 @@ def test_cancel_occurrence_notification(
...
 ):
-    study_groups = StudyGroupFactory.create_batch(3)
-    for s in study_groups:
-        EnrolmentFactory(study_group=s, occurrence=occurrence, person=s.person)
+
+    for status in Enrolment.STATUSES:
+        for s in StudyGroupFactory.create_batch(2):
+            EnrolmentFactory(
+                study_group=s, occurrence=occurrence, person=s.person, status=status
+            )
```
nikomakela, line 216: Create 2 enrolments with each status code.  Add the enrolments to occurrence. Enrolment needs an unique study group.

occurrences/tests/test_notifications.py
```diff
@@ -208,11 +208,21 @@ def test_cancel_occurrence_notification(
...
+    notifiable_enrolments_count = occurrence.enrolments.all().count() - (
+        occurrence.enrolments.filter(status=Enrolment.STATUS_CANCELLED)
+        .filter(status=Enrolment.STATUS_DECLINED)
+        .count()
+    )
```
nikomakela, line 221: Notification should be sent to contact persons who have not cancelled their enrolment yet or it has not declined.

occurrences/models.py
t0mim: Should we explicitly say which statuses are ok rather than saying which things are not ok? Just thinking that if someone adds a new status then its by default ok.

nikomakela: I was thinking the opposite :smile: I thought it would be better to send the notification by default. And there is a greater chance that the new status is one of the “allowed” ones rather than rejected ones.

Too much communication would be better than too little.

PR #171, opened 2021-05-20
Sending a no-cache cookie with rest queries to LE.

nikomakela:
PT-1043. Needed to add a cookie implementation to disable cache on certain calls.


nikomakela:
The fix is greatly needed for production use so I self approve this.

PR #173, opened 2021-05-21
Added a type_id to event schema with "general" as default value.

nikomakela:
PT-1044.


nikomakela:
I tried to make this field enum, but couldn't find a way to give a default value for it.
See:
https://github.com/graphql-python/graphene/issues/877
https://github.com/graphql-python/graphene/issues/691
https://docs.graphene-python.org/en/latest/types/enums/

nikomakela:
Closing the pull request because the wanted feature is not available yet: https://github.com/graphql-python/graphene/commit/c08379ed85b2759de32777eb5dd3dca143a6d69f

PR #179, opened 2021-06-20
Fixes to CSV file when reading it with Microsoft Excel

nikomakela:
PT-1052. Set BOM and semicolon as a delimiter to fix the encoding issues.

Screenshot from Microsoft Excel for Mac:
image

FYI: 
- https://docs.python.org/3/library/csv.html
- https://stackoverflow.com/questions/5202648/adding-bom-unicode-signature-while-saving-file-in-python
- https://theonemanitdepartment.wordpress.com/2014/12/15/the-absolute-minimum-everyone-working-with-data-absolutely-positively-must-know-about-file-types-encoding-delimiters-and-data-types-no-excuses/
- https://en.wikipedia.org/wiki/Comma-separated_values

### DIFF
--- a/reports/views.py
+++ b/reports/views.py
@@ -220,14 +220,44 @@ class ExportReportCsvView(ExportReportViewMixin, APIView):
     authentication_classes = [ApiTokenAuthentication, SessionAuthentication]
     permission_classes = [IsAdminUser]
     csv_dialect = csv.excel
+    csv_delimiter = ";"
 
     def _create_csv_response_writer(self, filename):
-        response = HttpResponse(content_type="text/csv; text/csv; charset=utf-8")
+        response = HttpResponse(content_type="text/csv; charset=utf-8")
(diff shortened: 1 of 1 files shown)


reports/views.py
```diff
@@ -219,16 +219,52 @@ class ExportReportCsvView(ExportReportViewMixin, APIView):
     model = None
     authentication_classes = [ApiTokenAuthentication, SessionAuthentication]
     permission_classes = [IsAdminUser]
+    csv_dialect = csv.excel
+    csv_delimiter = ";"
+
+    def _create_csv_response_writer(self, filename):
+        response = HttpResponse(content_type="text/csv; text/csv; charset=utf-8")
```
Sartonon, line 226: `text/csv;` needed twice?

Sartonon (approved):

reports/views.py
nikomakela: nope :) absolutely not

PR #180, opened 2021-07-01
Add model, schema, input and admin features for users organisation proposals

nikomakela:
PT-1071, PT-1074, PT-1083 API should be able to receive the 3rd party organisation name on user creation

When a member of a 3rd party organisation registers to the API from the providers UI, he can make a proposal to add a new 3rd party organisation.Since the process is still quite unclear, the proposals and and 3rd party organisation wishes can be collected with this model and stored in database as detached. The use case is that an admin can see which organisation the new user likes to represent and then the real organisation.Organisation instance can be created (if necessary and not yet done) and linked to the user.

Added an input for organisation proposals when creating a new profile. Also fixed an old implementation about users is_staff resolving.

PT-1076 As an admin I can view the 3rd party organisation name in admin form.

Show organisation proposals in user's admin change form. Some search improvements added to organisations admin. Added organisation proposals views to admin. 

createProfile mutation
image

query (maybe not ever needed)
image


Sartonon (approved):
Looks good to me!

nikomakela:
Something is still wrong with myProfile query done from palvelutarjotin-admin app. The tests goes through, but the returned profile is null (in my local env).

I also found another bug which does not select organisations in user's admin change form. They are properly selected in person's admin change form. The issue is not directly related to this ticket, but could be fixed in this PR.

PR #181, opened 2021-07-05
Send email to admins while new profile creations

nikomakela:
PT-1075 PT-1080. Send email to admins while new profile creations.

PT-1075 PT-1079. Added is_admin field to the model and admin. The is_admin field is used to identify the users to whom the notification is sent.

PT-1075 PT-1086. Added a "Can administrate user permissions" permission. The new can_administrate_user_permissions can be used to restrict the administrative actions done to user permissions created from the user change admin form.

NOTE: This PR needs https://github.com/City-of-Helsinki/palvelutarjotin/pull/180 as a base.

Hide for superuser:
image

Show fieldsets for superuser:
image

Readonly fields for users without the new permission:
image

PR #182, opened 2021-07-06
A notification is sent to a person when his account is accepted

nikomakela:
PT-1077. When a user has the is_staff -flag set to True and the person is linked to some organisations, when the user is updated from the user admin form, a notification will be sent.

Needs #180  and #181 as a base.

### DIFF
--- a/organisations/admin.py
+++ b/organisations/admin.py
@@ -1,5 +1,5 @@
 from django import forms
-from django.contrib import admin
+from django.contrib import admin, messages
 from django.contrib.admin.widgets import FilteredSelectMultiple
 from django.contrib.auth import get_user_model
 from django.contrib.auth.admin import UserAdmin as DjangoUserAdmin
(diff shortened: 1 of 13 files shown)


organisations/admin.py
```diff
@@ -198,6 +198,38 @@ class UserAdmin(DjangoUserAdmin):
...
+        """
+        Given a ModelForm return an unsaved instance. ``change`` is True if
+        the object is being changed, and False if it's being added.
+        Overridden so the request could be combined with the model instance save
+        and we could write a message about a mail is being sent.
+        """
+
+        original_user = form.instance
+
+        should_notify = self._has_organisations_changed(
```
Sartonon, line 211: maybe this variable name is a bit misleading if `should_notify = True` doesn't necessarily mean that there should actually be notification

Sartonon (approved):
Looks pretty good to me ;)

PR #189, opened 2021-07-12
Persons language is now editable in admin.

nikomakela:
PT-1090. Added a language field to person admin form.
Show only the users without  link to any person in person admin.

PT-1090
Added UserPersonInLine form to User admin change form.

PT-1090 Added person inline form in user change form, so it would be easier to find the person instance linked to the user.
Using admins and persons own language in my profile notifications.

PT-1090.
Changed the email is sent info message for admins.

PT-1096.

image

### DIFF
--- a/organisations/admin.py
+++ b/organisations/admin.py
@@ -1,4 +1,5 @@
 from django import forms
+from django.conf import settings
 from django.contrib import admin, messages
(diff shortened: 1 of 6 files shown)


Sartonon (approved):
Great :)

PR #194, opened 2021-08-20
Fixing issues when the link fails between palvelutarjotin event and linked event's event

nikomakela:
PT-1134 PT-1131. If LinkedEvents returns an event that does not match on any of our PalvelutarjotinEvent instance, an error was be raised. Now the resolve_events checks all the events in the response if they have a linked PalvelutarjotinEvent in our database.

Added some LinkedEvents API fail proof to occurrence signals.

### DIFF
--- a/graphene_linked_events/schema.py
+++ b/graphene_linked_events/schema.py
@@ -1,3 +1,6 @@
+import json
+from types import SimpleNamespace
+
 from django.core.exceptions import PermissionDenied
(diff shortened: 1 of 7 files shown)


nikomakela:
1 snapshot fails with empty event but it should have 1. Also, the new features should have some tests.

Sartonon (approved):

PR #195, opened 2021-09-02
Fixed enrolments counting.

nikomakela:
Fixed the calculation of seats left in occurrence.

PT-1137 PT-1143.

The issue was: "Hyväksymättä" for pending seats is always zero.
image


nikomakela:
Seems like the previous change to the seats calculation functions has been finished in a hurry :) The snapshot result has shown that it wasn't finished properly.

tuomas777 (approved):
LGTM 👍

PR #197, opened 2021-09-05
Events can be sorted and paginated by occurrences start time

nikomakela:
PT-1150 PT-1153.


nikomakela:
Sorting and pagination by occurrence data could be sorted out, but I think it won't be possible to include all the current filters in that query, because the filters are for events and the events are not stored in Kultus. So sorting and using pagination related to occurrences in Kultus DB and filtering used for events in LinkedEvents cannot be combined. 😞

nikomakela:
What was left to be done was filtering the results with LinkedEvents data. Also tests are missing.

PR #198, opened 2021-09-07
Added a neighborhood rest client and graphql schema.

nikomakela:
To offer a list of neighborhood divisions, the data needs to be fetched from kartta.hel.fi. The implementation is copied from Tapahtumat.hel.fi proxy, but it is ported from Node to Python

PT-1130 PT-1167.


neighborhood/rest_client.py
```diff
@@ -0,0 +1,34 @@
...
+
+    def neighborhood_list(self):
+        actions = self._get_actions(
+            "FeatureCollection", **{"typeName": "avoindata:Kaupunginosajako"}
+        )
+        return requests.request(
+            actions["list"]["method"],
+            actions["list"]["url"],
+            params=actions["list"]["params"],
+        )
```
tuomas777, line 34: How likely it is that there will be another action(s) added that will actually benefit from this `get_actions()` thingy, or is there some other reason for it? Right now this seems just like a very complex way to execute one request 🙂 I would just do the request in one simple method now, and add complexity later when/if needed.

organisations/services.py
```diff
@@ -30,6 +30,7 @@ def get_admin_emails_and_languages():
     return list(
         get_user_model()
         .objects.filter(is_admin=True)
+        .order_by("id")
```
tuomas777, line 33: This seems like a completely unrelated change, and therefore should be in its own commit IMO.

nikomakela: Some reason why the snapshot tests are sometimes rendered in a different order. Thats the reason for this :)

neighborhood/rest_client.py
nikomakela: Maybe you are right. I did it like this so it would work the same way as the linkedEvents rest_client does. But I don't see it would grow from this, so it could be simplified.

README.md
```diff
@@ -43,134 +45,141 @@ Create user and database
...
 
 Or you can use [`pre-commit`](https://pre-commit.com/) to quickly format your code before committing.
 
-
 1. Install `pre-commit` (there are many ways to do but let's use pip as an example):
-    * `pip install pre-commit`
+   - `pip install pre-commit`
 2. Set up git hooks from `.pre-commit-config.yaml`, run this command from project root:
-    * `pre-commit install`
+   - `pre-commit install`
```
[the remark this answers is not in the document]
nikomakela: Rebase-editted them to different commit.

organisations/services.py
nikomakela: Rebase-editted the change to a different commit.

neighborhood/rest_client.py
nikomakela: Simplified it by removing get_actions and the tests related to it.

tuomas777 (approved):
Splendid! 👍 🚀

PR #199, opened 2021-09-10
Null pointer checks added for location reading in notification templates

nikomakela:
PT-1166


nikomakela:
This has gone old. #250 contains the same fixes.

PR #201, opened 2021-09-16
Fix to events list meta count

nikomakela:
LinkedEvents paginates the results, but Kultus API filters the paginated sets. This leads to a situation where the events count in meta data, easily does not match with the fact. If events.meta is used, the count is bigger than the filtered result. If a length of  integrated events is used, it would work only for unpaginated results. Using meta given by LinkedEvents is better, but can lead to empty pages!

PT-1170

Only ways to get the pagination work every time would be 
1. unpaginated result set from LE
2. Kultus includes a list of every tested event ids to the query.

We cannot do either! However, this solution introduced by this Pull request is better alternative than the current active one.

### DIFF
--- a/graphene_linked_events/schema.py
+++ b/graphene_linked_events/schema.py
@@ -399,10 +399,13 @@ class Query:
 
         # Create a (new) mutated immutable X-object of the event results.
         return json_object_hook(
-            {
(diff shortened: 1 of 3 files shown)


Sartonon (approved):
Looks good!

PR #204, opened 2021-09-17
Added place_ids to Person to list persons own places they want to follow.

nikomakela:
Using a Postgre specific ArrayField to store an array of place ids (which are strings). The place model is defined in LinkedEvents, so the link to the model needs to be loosely set.

PT-1162 PT-1201


organisations/factories.py
```diff
@@ -3,6 +3,13 @@
 from organisations.models import Organisation, OrganisationProposal, Person
 
 
+def fake_id_array():
+    from faker import Faker
+
+    fake = Faker()
+    return fake.pylist(3, True, "str")
```
tuomas777, line 10: Hmm, what keeps these values the same between test runs (for snapshots)?

organisations/schema.py
```diff
@@ -46,6 +46,7 @@ class PersonNodeInput(InputObjectType):
     phone_number = graphene.String()
     email_address = graphene.String(required=True)
     language = LanguageEnum(description="Default `fi`")
+    place_ids = graphene.List(graphene.String, default=[])
```
tuomas777, line 49: The default value shouldn't be mutable. Does this actually need a default value?

organisations/tests/test_models.py
```diff
@@ -47,6 +47,59 @@ def test_membership_creation():
...
+    ]
+    assert list(Person.objects.filter(place_ids__overlap=[place1, place3])) == [
+        person1,
+        person2,
+        person3,
+    ]
+
+    assert list(Person.objects.filter(place_ids__len=1)) == [person2]
+
+
```
tuomas777, line 102: I don't quite understand the meaning of this test, I mean isn't this basically just testing that Django's `ArrayField` works? 😄

Also, there is now no test to verify the GraphQL mutation works, I would rather do that 🙂

organisations/factories.py
nikomakela: The seed configuration, I guess(?).
```
@pytest.fixture(autouse=True)
def setup_test_environment(settings):
    settings.CAPTCHA_ENABLED = False
    factory.random.reseed_random("777")
    with freeze_time("2020-01-04"):
        yield
```

I admit that I would need to get to know the subject a little more closely myself. I’m going to do it as soon as I get time on the subject. Sometimes we need to update snapshots that are not even remotely related to the updated code and I would like to know why. So far I have blamed it in part on Graphene’s fault

nikomakela: If it would change every time, Github would not pass my snapshot 😃

organisations/schema.py
nikomakela: you are absolutely right! No need as it is provided at the data model level

organisations/tests/test_models.py
nikomakela: first of all, I want it for myself! 😆  I know the field type is tested by Django already, but I really wanted to test its functionality, because I saw that it have had a bug in Django some time ago and it's a Postgre specific field. I originally wrote it while solving the problem with Api and speculating on whether the field needs to be forced to nullable.

So, if you don't mind, I leave it there. At least nobody should ever have to touch on it.

And yeah, I thought I missed something since this was a draft PR this morning. I'm going to add those mutation tests - and maybe even implement the feature first 😄

tuomas777 (approved):
Good job! 👍

PR #205, opened 2021-09-20
Fixes to event publishing when event has non-internal enrolments

nikomakela:
Updated event publishing to support nullable enrolment end days. Added new tests for publishing the event with different enrolment types. PT-1211.


Sartonon (approved):
Excellent

PR #208, opened 2021-09-23
Added a timeout setting to all external requests

nikomakela:
Without setting the timeout implicitly the timeout is set to None. Now it's set to 5 seconds on every requests that I found.

PT-1183.


nikomakela:
Tried first with a 5 seconds timeout, but it ended up to failing tests 😄 So, 10 secs it is.

tuomas777 (approved):
👍

PR #209, opened 2021-09-27
Notification template importers

nikomakela:
The base idea is copied from Kukkuu project (https://github.com/City-of-Helsinki/kukkuu/tree/master/importers).
The implementation from Kukkuu is converted to an abstract that is then used to implement 2 different kind of importers: 1 that imports from Google Sheet like Kukkuu does and one that uses template files from the project itself.

----

The notification templates can be imported via
    a. a Google sheet importer
    b. a Template file importer

The importer can be used to create and update the notification templates or to check whether the templates are in sync.
The importer can be used via Django management commands (in notification_importers app) or admin site tools.

To enable admin site tools, some configuration is needed:

To enable a selected importer (`NotificationFileImporter` or `NotificationGoogleSheetImporter`)

```python
NOTIFICATIONS_IMPORTER = (
    "notification_importers.notification_importer.NotificationFileImporter"
)
```

If a Google sheet importer is used, also `NOTIFICATIONS_SHEET_ID` is needed

```python
NOTIFICATIONS_SHEET_ID = "1234"
```

If a File importer is used, files should be stored in notification_importers app in a notification_importers/templates/sms and notification_importers/templates/email folders.

There is also a naming convention used there. The file name must be given in this pattern [notification_type]-[locale].[html|j2].

image


tuomas777 (approved):
great

PR #214, opened 2021-10-08
Fix to reports views OIDC-authentication

nikomakela:
Reports view needed a fix for AMR-JWT-claim, which was returned as a string instead of a list of strings by Tunnistamo. Same fix was already done to GraphQL requests.

PT-1218.

### DIFF
--- a/palvelutarjotin/oidc.py
+++ b/palvelutarjotin/oidc.py
@@ -1,16 +1,7 @@
 from helusers.oidc import ApiTokenAuthentication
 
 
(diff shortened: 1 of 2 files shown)


charn (approved):
The real problem imho seems to be in Tunnistamo, which as you pointed out, returns the `amr` claim as a string instead of as a list of strings (which is should be by JWT specification).

This fix works. :+1:  One other option would be to use `drf-oidc-auth==0.10.0`. Later versions of that library changed from `jwkest` to `authlib` which checks for correctly formatted `amr` claim.

PR #219, opened 2021-11-01
Start time and enrolment times are set from occurrences when event is published

nikomakela:
PT-1256:
- Removed the CourseExtension from Kultus API, because it has been deleted from LinkedEvents API since LinkedEvents v2.
- Enrolment start and end times are published and updated to LinkedEvents with event start and end times.

When published:

1. The event start time is the start time of the event's first occurrence by the occurrences start times.

2. The event end time is the end time of the event's last occurrence by the occurrences end times.

3. Enrolment start time is the PalvelutarjotinEvent start time. If there are no occurrences (like in UI's 1st step), the start time will be current time!

4. Enrolment end time is the start time of the event's last occurrence by the occurrences start times subtracted with the value of enrolment_end_days field of PalvelutarjotinEvent.


occurrences/event_api_services.py
```diff
@@ -105,3 +119,36 @@ def get_event_time_range_from_occurrences(p_event: Optional["PalvelutarjotinEven
...
+        1. No p_event set yet
+        2. 2nd step of the UI's form wizard, that includes the occurrences
+        and enrolment data, is not yet full filled.
+        3. The event is externally enrollable
+        4. The event is not enrollable at all
+        """
+        return None, None
+
+    occurrences = p_event.occurrences.filter(cancelled=False).order_by("end_time")
+    end_time = occurrences.last().start_time
```
tuomas777, line 138: This raises if there aren't any occurrences, but I suppose we can assume there always are at this point?

occurrences/tests/test_event_api_services.py
```diff
@@ -15,15 +16,34 @@ def test_send_event_republish(
     mock_update_event_data,
     p_event,
 ):
-    OccurrenceFactory(p_event=p_event)
+    OccurrenceFactory.create_batch(3, p_event=p_event, cancelled=False)
+
+    first_occurrence = p_event.occurrences.order_by("start_time").first()
+    last_occurrence = p_event.occurrences.order_by("end_time").last()
```
tuomas777, line 22: You could also use little more succinct versions `.earliest("start_time")` and `.latest("end_time")`.

occurrences/tests/test_event_api_services.py
```diff
@@ -15,15 +16,34 @@ def test_send_event_republish(
...
-        p_event.get_enrolment_end_time_from_occurrences()
+    assert event_obj["start_time"] == format_linked_event_datetime(event_start)
+    assert event_obj["end_time"] == format_linked_event_datetime(event_end)
+    assert event_obj["enrolment_start_time"] == format_linked_event_datetime(
+        p_event.enrolment_start
+    )
+    assert event_obj["enrolment_end_time"] == format_linked_event_datetime(
+        enrolling_ends
     )
+    assert event_obj["enrolment_end_time"] is not None
```
tuomas777, line 46: This seems like an unnecessary check as `format_linked_event_datetime` does not seem to allow or return null values. Then again, I think you could consider changing it to actually accept null values to get rid of tons of those checks like `x = format_linked_event_datetime(dt) if dt else None` 🙂

nikomakela: In that test case I really want to check that it is not None

nikomakela: I'll still add a null check to that `format_linked_event_datetime` -function.

graphene_linked_events/tests/test_api.py
```diff
@@ -1278,16 +1276,33 @@ def test_publish_event_unauthorized(
...
     snapshot,
     api_client,
     user_api_client,
     staff_api_client,
     mock_get_draft_event_data,
-    mock_update_event_data,
     organisation,
     person,
 ):
+    def __eq_dt_with_tz(dt1, dt2) -> bool:
```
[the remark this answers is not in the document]
nikomakela: It's not used anywhere else so I don't want it to be very global. I can still add it to the top of `test_api.py`.

charn (approved):
LGTM! :+1: :shipit:

PR #220, opened 2021-11-01
API serves a list of schools and kindergartens of City of Helsinki.

nikomakela:
Added a GraphQL API schema that can be used to fetch a list of ids' and names of schools, colleges and kindergartens of Helsinki via Servicemap RESTclient (https://api.hel.fi/servicemap/v2/unit/?municipality=helsinki&only=name&service_node=2118%2C868%2C2179%2C1088%2C1257%2C1097).

PT-1234 PT-1264.


nikomakela:
image

Hmm.. seems like this needs some charset handling

servicemap/rest_client.py
```diff
@@ -0,0 +1,44 @@
+import requests
+
+
+class ServicemapApiClient:
+
+    CONNECTION_TIMEOUT = 20
+    DATA_SOURCE = "tprek"
```
nikomakela, line 7: Is the "tprek" the only datasource here? I set the id to be "tprek:{ID}" so that it would be in same format as it is in the Kultus database, in LinkedEvents API and in rest of the system.

servicemap/schema.py
```diff
@@ -0,0 +1,52 @@
...
+    schools_and_kindergartens_list = Field(
+        ServiceUnitNameListResponse, page=Int(), page_size=Int()
+    )
+
+    @staticmethod
+    def normalize_service_unit_node(units: list):
+
+        return [
+            {
+                "id": "{data_source}:{place_id}".format(
```
nikomakela, line 25: Is the "tprek" the only datasource here? I set the id to be "tprek:{ID}" so that it would be in same format as it is in the Kultus database, in LinkedEvents API and in rest of the system.

servicemap/rest_client.py
```diff
@@ -0,0 +1,44 @@
...
+
+    @staticmethod
+    def __convert_to_string_param(params):
+        if not params:
+            return None
+        for k, v in params.items():
+            if type(v) == list:
+                list_to_string = ",".join(v)
+                params[k] = list_to_string
+        return params
```
tuomas777, line 44: While it probably doesn't cause actual problems with the current code, I don't think it is optimal that this function, and therefore the outer `list_helsinki_schools_and_kindergertens()` as well, actually modify the list in `filter_list`, as callers most probably don't except that, and there is also no real need for it.

tuomas777 (approved):
Added one minorish comment, other than that LGTM 👍 🏫

palvelutarjotin/settings.py
```diff
@@ -294,6 +295,8 @@
     "DATA_SOURCE": env.str("LINKED_EVENTS_DATA_SOURCE"),
 }
 
+SERVICEMAP_API_CONFIG = {"ROOT": env.str("SERVICEMAP_API_ROOT")}
```
charn, line 298: Just FYI, since you've defined `SERVICEMAP_API_ROOT` as a string, you can here also just use `env("SERVICEMAP_API_ROOT")`

servicemap/rest_client.py
```diff
@@ -0,0 +1,45 @@
...
+    def __init__(self, config) -> None:
+        self.root = config["ROOT"]
+
+    def list_helsinki_schools_and_kindergartens(self, filter_list=None):
+        """
+        List of names and ids' of schools, colleges and kindergardens.
+        Query:
+        https://api.hel.fi/servicemap/v2/unit/?municipality=helsinki&only=name&service_node=2118,868,2179,1088,1257,1097
+        """
+        service_nodes = [2118, 868, 2179, 1088, 1257, 1097]
```
charn, line 18: I wonder if it would make sense to have this in settings e.g. in the `SERVICEMAP_API_CONFIG`

nikomakela: I thought so first too... It could, but the service nodes should be VERY static and should only affect in this 1 specific API call. In that sense, it 's nothing but more things to configure.

setup.cfg
```diff
@@ -1,5 +1,5 @@
 [flake8]
-exclude = migrations,snapshots
+exclude = migrations,snapshots,servicemap/tests/mock_data.py
```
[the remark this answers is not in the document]
nikomakela: I'll change it to use what you advised.

servicemap/rest_client.py
nikomakela: I will add comments about what they mean next to those service node codes.

nikomakela:
This new API end point would actually need a search input filter also, because over a thousand places will be returned. The problem is that service map API does not offer search input parameter in their unit-list-endpoint. They do have it in their search-endpoint, but there is no way to filter with service nodes, but only with services.

This needs to be somehow settled! But it could be done in another PR.

PR #221, opened 2021-11-04
Replace Study group name field with a new unit field

nikomakela:
Study group handles the new unit id field along with renamed unit name field.

The old name field of StudyGroup model is renamed to "unit_name". A new "unit_id" field is added alongside the unit name field. The new unit id can be used to fetch a whole Place object from external APIs like LinkedEvents or Servicemap.

PT-1236 PT-1264 PT-1265

StudyGroup schema's unit field is now working with an union type: the unit can be a Place or a new ExternalPlace, which includes the name only.

NOTE: There were a lot of issues with circular imports, which are now being fixed with strings instead of real package imports.

To offer some backward compatibility for notification templates etc., there is a name getter that return unit name

PT-1264.

----

The new query looks like this:

```
query {
  studyGroups {
    edges {
      node {
        unitName
        unitId
        unit {
          ... on ExternalPlace {
            name {
              fi
              sv
            }
          }
          ... on Place {
            internalId
            id
            name {
              fi
              sv
              en
            }
            position {
              coordinates
            }
          }
        }
      }
    }
  }
}

```

The new variables:

```
ADD_STUDY_GROUP_VARIABLES = {
    "input": {
        "person": {
            "name": "Name",
            "emailAddress": "email@address.com",
            "phoneNumber": "123123",
            "language": "SV",
        },
        "unitId": "tprek:123",
        "unitName": "Sample study group name",
        "groupSize": 20,
        "amountOfAdult": 1,
        "studyLevels": ["GRADE_1"],
        "groupName": "Sample group name",
        "extraNeeds": "Extra needs",
    }
}
```


nikomakela:
NOTE: this branch will be merged to PT-1264/servicemap-api

graphene_linked_events/schema.py
```diff
@@ -215,8 +210,8 @@ class Event(IdObject):
     info_url = Field(LocalisedObject)
     provider_contact_info = String()
     description = Field(LocalisedObject, required=True)
-    p_event = Field(PalvelutarjotinEventNode, required=True)
-    venue = Field(VenueNode)
+    p_event = Field("occurrences.schema.PalvelutarjotinEventNode", required=True)
```
nikomakela, line 213: Preventing circular imports...

occurrences/models.py
```diff
@@ -427,7 +428,13 @@ class StudyGroup(TimestampedModel):
     person = models.ForeignKey(
         "organisations.Person", verbose_name=_("person"), on_delete=models.PROTECT
     )
-    name = models.CharField(max_length=1000, blank=True, verbose_name=_("name"))
+    # Tprek / Service map id for school or kindergarten from the city of Helsinki
+    unit_id = models.CharField(
+        max_length=255, verbose_name=_("unit id"), blank=True, null=True
```
charn, line 433: What does it mean for us if `unit_id` is `None` instead of an empty string i.e. should we just remove `null=True`?

nikomakela: `None` represents the situation when the value is really not given. In this specific case it means that unit, a school or kindergarten, is outside of the city of Helsinki. The unit_id -field stores a loose id which can be linked to a TPREK service unit (in the Servicemap or the LinkedEvents). If the unit_id is not given, the school is outside from the city of Helsinki and only the name should and must be given.

I would say that from the data and database perspective, it's much more clear and therefore an important thing to store the value as NULL, when the value is unknown or not given. Me myself, I would not allow the `blank=True` here, or almost never, because an empty string really does not represent a value that is not given, but rather a value which would like to be bypassed in a form somehow. I think that setting the field to both, nullable and blankable is a preferred way in Django or at least in Palvelutarjotin, because the model file is full of these:

```
    enrolment_start = models.DateTimeField(
        verbose_name=_("enrolment start"), blank=True, null=True
    )
    # Enrolment will be close x days before the occurrence start
    enrolment_end_days = models.PositiveSmallIntegerField(
        verbose_name=_("enrolment end days"), blank=True, null=True
    )
    external_enrolment_url = models.URLField(
        verbose_name=_("enrolment url"), blank=True, null=True
    )
    needed_occurrences = models.PositiveSmallIntegerField(
        verbose_name=_("needed occurrence"), default=1
    )
```

If I'm to decide, I would set them all to `null=True, blank= False`. Is the blank-attribute here to make the validation simpler or just to fool Django somehow?

nikomakela: To me it's irritating that I need to check things like this https://github.com/City-of-Helsinki/palvelutarjotin/pull/221/files#diff-32cfd933813e52d2e31fe5b92c696bb856c1d524a38666af02944ce964511930R40. Therefore I would not allow them to be blank, but only None.

charn: `CharField`s are a bit different from the other field types since empty string usually can be considered as empty, where as there's no such value for datetimes. Having `blank=False` (which is the default) makes sure that if there's any type of form validation happening, it will not allow an empty value on that field. I'm not sure how the validations are being run for graphene, but at least normal Django forms (including the ones in Django admin) and Django Rest framework serializers run the validations.

But I guess since there's a semantic difference here between `None` and `""` it's probably ok to have `null=True` here.

charn (approved):
LGTM! :+1: :shipit:

PR #222, opened 2021-11-04
PT-1267 | Add popular_kultus_keywords GQL query

charn:
Kultus related categories are fetched from Linked Events and returned sorted by the number of events.

Linked Events does not currently provide filters which would allow fetching relevant keyword e.g. filtering based on the keyword set and then sorting by the number of events.


graphene_linked_events/schema.py
```diff
@@ -472,6 +479,32 @@ def resolve_keywords(parent, info, **kwargs):
...
+        keywords = []
+
+        for set_id in set_ids:
+            response = api_client.retrieve(
+                "keyword_set", set_id, params={"include": "keywords"}
+            )
+            response = json2obj(format_response(response))
+            keywords += response.keywords
+
+        keywords = sorted(keywords, key=lambda x: x.n_events, reverse=True)
```
nikomakela, line 505: Duplicates should be filtered out. It's not an issue in a staging or production environment, but it is in my local :) 
Maybe with:
```
unique_keywords = list(set(keywords))
keywords = sorted(unique_keywords, key=lambda x: x.n_events, reverse=True)
```

graphene_linked_events/schema.py
```diff
@@ -367,6 +369,11 @@ class Query:
     )
     keyword = Field(Keyword, id=ID(required=True))
 
+    popular_kultus_keywords = Field(
+        KeywordListResponse,
+        description=_("Keywords related to Kultus ordered by the number of events."),
```
nikomakela, line 374: Good!

nikomakela (changes requested):
A trivial thing and maybe even some over engineering, but I would add an amount-filter and maybe an asc/desc/ -sorter. Duplicates should be filtered out anyway. And at least 1 simple snapshot test should be added.

nikomakela (approved):
Great!

PR #223, opened 2021-11-11
Recursive rest-client call to fetch all schools and kindergarten pages at once

nikomakela:
The service map API response with 1000+ schools and kindergartens, but the max page size is 1000. Because the service map's unit query does not support filtering with a search term and servicemap's search -query does not support filttering with service nodes, we needed a solution to return all the units (1000+) in a single page all at once. This is now achieved by returning all the pages from graphql API (with schoosAndKindergartensList -query). The next page is called recursively when one is available. 

PT-1234 PT-1298 PT-1312.

### DIFF
--- a/servicemap/schema.py
+++ b/servicemap/schema.py
@@ -1,5 +1,5 @@
 import graphene
-from servicemap.utils import api_client, json2obj, json_object_hook
+from servicemap.utils import api_client, get_params_from_url, json2obj, json_object_hook
 
 
 class Response(graphene.ObjectType):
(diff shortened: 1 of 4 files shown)


nikomakela:
- [ ] Tests are still to come!

servicemap/schema.py
```diff
@@ -40,16 +38,35 @@ def normalize_service_unit_node(units: list):
...
+            # Because all pages are combined,
+            # the next_page_url should be set to None
+            # after the last response is retrieved.
+            next_page_url = None
+
         result = {
             "meta": {
                 "count": json_data.count,
-                "next": json_data.next,
+                "next": next_page_url,
```
charn, line 66: If my brain is working correctly this morning, I think you should be able to just do `"next": None` here. :thinking:

charn (approved):
LGTM! :+1: :shipit:

PR #225, opened 2021-11-16
PT-1272 | Add GraphQL query for fetching upcoming events

charn:
This PR adds a query for fetching upcoming events.


nikomakela (approved):
Nice! Great stuff with p_events query! The test is effective but a bit hard to understand.

PR #226, opened 2021-11-16
Added the EnrolmentReport model

nikomakela:
PT-1235 PT-1303 PT-1307.


reports/models.py
```diff
@@ -0,0 +1,175 @@
...
+    keep some information so the data is still available for reports:
+        1. Group sizes: Kids and adults
+        2. Study levels
+    - Store the unit position (coordinates) and divisions to normalise the data.
+    Place data must be fetched from LinkedEvents API
+    (or from other service like the servicemap API)
+    """
+
+    study_group = models.ForeignKey(
+        occurrences_models.StudyGroup, on_delete=models.DO_NOTHING,
```
charn, line 36: Maybe DO_NOTHING could be replaced with `null=True, on_delete=models.SET_NULL`

reports/tests/test_utils.py
```diff
@@ -0,0 +1,26 @@
...
+# @pytest.mark.parametrize(
+#     "lon1,lat1,lon2,lat2,distance_km",
+#     [
+#         [24.9384, 60.1699, 22.2666, 60.4518, 150.44]
+#     ],  # Helsinki  # Turku  # Distance in km
+# )
+# def test_get_distance(
+#     lon1: float, lat1: float, lon2: float, lat2: float, distance_km: float
+# ):
+#     assert get_distance(lon1, lat1, lon2, lat2) == distance_km
```
nikomakela, line 14: I have no idea why this gives results something like 250km (Helsinki - Turku). Can the 3D calculator make so much difference? It seems like the Google Map is using 2D and the same haversine -algorithm (great circle). I think it's ok to use it and did not need any external library

nikomakela:
Ready for final review
- [ ] There are couple of Django TestCase unit tests parametrized with pytest.mark.parametrize which does not work like that. Need to fix them - They are tested manually anyway.

occurrences/migrations/0035_add_enrolment_updated_at_field.py
```diff
@@ -0,0 +1,33 @@
+# Generated by Django 3.2.5 on 2021-11-17 19:48
+
+from django.db import migrations, models
+from django.db.models import F
+
+
+def set_enrolment_time_tom_updated_at(apps, schema_editor):
```
charn, line 7: Typo in the function name. :eyes:

occurrences/migrations/0035_add_enrolment_updated_at_field.py
```diff
@@ -0,0 +1,33 @@
...
+            field=models.DateTimeField(auto_now=True, verbose_name="updated at"),
+        ),
+        migrations.AlterField(
+            model_name="enrolment",
+            name="enrolment_time",
+            field=models.DateTimeField(
+                auto_now_add=True, verbose_name="enrolment time"
+            ),
+        ),
+        migrations.RunPython(set_enrolment_time_tom_updated_at),
```
charn, line 32: Not really required, but this could be:
```suggestion
        migrations.RunPython(set_enrolment_time_tom_updated_at, migrations.RunPython.noop),
```

reports/management/commands/sync_enrolment_reports.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+class Command(BaseCommand):
+    help = (
+        "Synchronize enrolment reports from enrolment table. "
+        + "Latest sync time will be determined from "
+        + "updated_at -field in a enrolment_reports table. "
+        + "Enrolments after that date will be updated "
+        + "if the enrolment status has changed "
+        + "and created if they are totally missing from the table content."
```
charn, line 14: AFAIK, since you have parentheses here you can also do this without the plus signs.
```suggestion
        "Synchronize enrolment reports from enrolment table. "
        "Latest sync time will be determined from "
        "updated_at -field in a enrolment_reports table. "
        "Enrolments after that date will be updated "
        "if the enrolment status has changed "
        "and created if they are totally missing from the table content."
```

reports/management/commands/sync_enrolment_reports.py
```diff
@@ -0,0 +1,42 @@
...
+        parser.add_argument(
+            "--ignore_linkedevents",
+            action="store_true",
+            help="Determine the date from where to start the sync",
+        )
+
+    def handle(self, *args, **options):
+        sync_from = None
+        if options["sync_from"]:
+            sync_from = datetime.fromisoformat(options["sync_from"][0])
```
charn, line 34: Django also has `parse_datetime` in `django.utils.dateparse`. Not sure if it has any added benefits thought.

reports/views.py
```diff
@@ -350,3 +357,52 @@ def get(self, request, *args, **kwargs):
...
+def sync_enrolment_reports_view(request):
+    create_from = Enrolment.objects.aggregate(Min("enrolment_time"))[
+        "enrolment_time__min"
+    ]
+    sync_enrolment_reports(hydrate_linkedevents_event=True, create_from=create_from)
+    messages.add_message(request, messages.SUCCESS, _("Enrolment reports synced!"))
+    return HttpResponseRedirect(reverse("admin:reports_enrolmentreport_changelist"))
+
+
+class EnrolmentReportListView(generics.ListAPIView):
```
charn, line 377: I'm guessing that the Power Bi user will use either `SessionAuthentication` or `BasicAuthentication`?

reports/management/commands/sync_enrolment_reports.py
nikomakela: okay, didn't know. It sure looks better

reports/management/commands/sync_enrolment_reports.py
nikomakela: I'll try it later. Had to fight a bit to find the solution with datetime.fromisoformat, but now it seems to work ok! 😄

reports/views.py
nikomakela: Yep. So do need any changes here? Do we need to narrow the selection?

charn: Just confirming that it was the plan. No need to change anything. :+1:

nikomakela:
Added some documentation to enrolment report list view
image

reports/admin.py
```diff
@@ -0,0 +1,293 @@
...
+
+    get_sync_status.short_description = _("in sync with the enrolment")
+    get_sync_status.boolean = True
+
+    def get_study_level_labels(self, obj):
+        return ",".join(
+            [dict(enumerate(sl)).get(1, "") for sl in obj.study_group_study_levels]
+        )
+
+    get_study_level_labels.short_description = _("Study levels")
```
charn, line 213: Just FYI, these days you can replace this with `@admin.display(description=_("Study levels"))`

reports/admin.py
```diff
@@ -0,0 +1,293 @@
...
+            )
+            self.message_user(request, message, messages.WARNING)
+        return super().changelist_view(request, *args, **kwargs)
+
+    def get_sync_status(self, obj):
+        # TODO: Cache needed? - May be too heavy to load for each
+        return obj._enrolment.status == obj.enrolment_status
+
+    get_sync_status.short_description = _("in sync with the enrolment")
+    get_sync_status.boolean = True
```
charn, line 206: You could also use `@admin.display(description=_("in sync with the enrolment"), boolean=True)`

reports/utils.py
```diff
@@ -0,0 +1,34 @@
+from math import asin, cos, radians, sin, sqrt
+
+LINKED_EVENTS_GEOS_PROJECTION_SRID = 3067
+
+"""
+FIXME: get_distance would be better alternative to haversine(),
+since it calculates the 3d distance.
+Can the 3d distance really be twice the 2d distance or why is the number so much bigger?
+"""
+# from django.contrib.gis.measure import Distance
```
charn, line 10: As pointed out earlier, this would need gdal libraries. So you might neet to do sometthing like this:
- https://github.com/City-of-Helsinki/maritime-maas/blob/main/.github/workflows/ci.yml#L58
- https://github.com/City-of-Helsinki/maritime-maas/blob/main/Dockerfile#L13-L14

reports/admin.py
nikomakela: Good to know in future. They could all be changed all at once in another ticket.

reports/admin.py
nikomakela: Good to know in future. They could all be changed all at once in another ticket.

reports/models.py
```diff
@@ -0,0 +1,468 @@
...
+    """
+    Enrolment report model is used to offer some exportable data related to enrolments.
+    It stores the information about enrolments and their study groups,
+    the coordinates of the group unit position and the coordinates of
+    the event position to calculate the distance to event,
+    the event data like time frame of the event and te publisher,
+    capacity information, event keywords, etc.
+    """
+
+    """
```
charn, line 148: Just writing about my personal preferences here. I'm not entirely sure if there's is a strong convention regarding block comments. There's [this](https://www.python.org/dev/peps/pep-0008/#block-comments), but I'm not sure if that's a strong enough argument. I've usually only seen multi-line string used for strings and docstrings. For one or multi-line comments I personally always use `#`.

I'm not entirely sure if the Python will evaluate this string or would it be able to optimize and ignore it like comments beginning with `#`. :thinking:

nikomakela: I changed them. Here is a conversation about them https://stackoverflow.com/a/7696966/784642

reports/utils.py
nikomakela: Too much for too little at this time, I think. We could consider this later, but now I self prefer that we go with 2d model. The difference won't even matter in our use.

nikomakela: But still, nice to know how to get it to work.

reports/models.py
```diff
@@ -0,0 +1,459 @@
...
+from django.utils.translation import ugettext_lazy as _
+from reports.exceptions import EnrolmentReportCouldNotHydrateLinkedEventsData
+from reports.utils import haversine
+
+from common.models import TimestampedModel
+
+logger = logging.getLogger(__name__)
+
+
+class EnrolmentQuerySet(models.QuerySet):
```
charn, line 18: I find it a bit confusing that we have a a manager/queryset on a model that returns a completely different model instances than what the `self.model` in this case would be. Also there's maybe some hacky way of using self here.

Maybe we could instead just have this functionality as functions? 

```python
def get_unsynched_enrollments(
    sync_from: datetime = None,
) -> Union[models.QuerySet, List[occurrences_models.Enrolment]]:
    """Get a list of enrolments which are updated after latest report updates."""
    sync_from = sync_from or EnrolmentReport.objects.latest_sync()
    if sync_from:
        return occurrences_models.Enrolment.objects.filter(updated_at__gte=sync_from)
    return occurrences_models.Enrolment.objects.all()


def get_missing_enrollments(
    sync_from: datetime = None,
) -> Union[models.QuerySet, List[occurrences_models.Enrolment]]:
    """Get a list of enrolments which are missing from EnrolmentReport db-table."""
    enrolments = get_unsynched_enrollments(sync_from=sync_from)
    reports_enrolment_ids = EnrolmentReport.objects.filter(
        _enrolment_id__in=[e.id for e in enrolments]
    ).values_list("_enrolment_id", flat=True)
    return enrolments.exclude(id__in=reports_enrolment_ids)
```

reports/models.py
```diff
@@ -0,0 +1,459 @@
...
+        Enrolment report is not in sync when it's related enrolment instance
+        has an updated_at -field value greater than
+        report instances updated_at -field value.
+        """
+        # Even if the enrolment instance has updated,
+        # we do not want to update it's related enrolment report instance,
+        # until the status of enrolment has changed.
+        enrolment_ids_with_statuses = list(
+            {"_enrolment_id": id, "enrolment_status": status}
+            for id, status in EnrolmentQuerySet.unsynced(self).values_list(
```
charn, line 58: If `EnrolmentQuerySet` is turned into functions this could be:

```suggestion
            for id, status in get_unsynched_enrollments().values_list(
```

reports/models.py
```diff
@@ -0,0 +1,459 @@
...
+        return (
+            occurrences_models.Enrolment.objects.all().count()
+            - self.model.objects.all().exclude(_enrolment__isnull=True).count()
+        )
+
+    def create_missing(
+        self, hydrate_linkedevents_event=False, sync_from: datetime = None
+    ):
+        """Create missing enrolment report instances."""
+        enrolments = self.model.enrolment_objects.missing(sync_from=sync_from)
```
charn, line 118: If `EnrolmentQuerySet` is turned into functions this could be:

```suggestion
        enrolments = get_missing_enrollments(sync_from=sync_from)
```

reports/models.py
```diff
@@ -0,0 +1,459 @@
...
+                ),
+            ).exclude(
+                enrolment_id_status__in=[
+                    "{}{}{}".format(
+                        entry["_enrolment_id"], separator, entry["enrolment_status"]
+                    )
+                    for entry in enrolment_ids_with_statuses
+                ]
+            )
+        )
```
charn, line 89: I might be totally wrong on this, but you might be able to shorten this to:
```python
        enrollments = get_unsynched_enrollments()
        return self.filter(_enrolment__in=enrollments).exclude(
            enrolment_status=F("_enrolment__status")
        )
```

reports/models.py
```diff
@@ -0,0 +1,459 @@
...
+
+    # Enrolment
+    # - Store the foreign key to fetch all data
+    # - If the foreign key integrity is gone,
+    # keep some information so the data is still available for reports:
+    #     1. enrolment time
+    #     2. enrolment status
+    # - Enrolment status is also used to check whether or not
+    # the db row should be updated.
+    _enrolment = models.ForeignKey(
```
charn, line 201: If there should every only be one report line per enrolment, we could make this `OneToOneField`, which would add the appropriate DB restrictions.

reports/models.py
```diff
@@ -0,0 +1,402 @@
...
+    # - Store the foreign key to fetch all data
+    # - If the foreign key integrity is gone,
+    # keep some information so the data is still available for reports:
+    #     1. Amount of the seats
+    #     2. cancelled status
+    # - Store the occurrence languages
+    # - Store the event position (coordinates) and divisions to normalise the data.
+    # Place data must be fetched from LinkedEvents API
+    # (or from other service like the servicemap API)
+    _occurrence = models.OneToOneField(
```
charn, line 166: I think only `_enrolment` can be `OneToOneField` since a single occurrence can have multiple enrolments? Same probably applies to the `_study_group` as well.

reports/models.py
```diff
@@ -0,0 +1,402 @@
...
+        max_digits=9,
+        decimal_places=2,
+        verbose_name=_("distance from unit to event place"),
+        null=True,
+    )
+
+    objects = EnrolmentReportManager()
+
+    @property
+    def study_group_group_size(self):
```
charn, line 251: Is this `study_group_group_size` property used somewhere? Should it also include the amount of adults?

reports/tests/test_utils.py
```diff
@@ -0,0 +1,26 @@
...
+@pytest.mark.parametrize(
+    "lon1,lat1,lon2,lat2,distance_km",
+    [
+        [24.9384, 60.1699, 22.2666, 60.4518, 150.4375814968687]
+    ],  # From Helsinki to Turku, it's 150.4km
+)
+def test_haversine(
+    lon1: float, lat1: float, lon2: float, lat2: float, distance_km: float
+):
+    assert haversine(lon1, lat1, lon2, lat2) == distance_km
```
tuomas777, line 26: Checking float equality is not a good idea. There is [math.isclose()](https://docs.python.org/3/library/math.html#math.isclose) for example that could be used instead.

reports/models.py
nikomakela: Changed them to be onetoones

reports/models.py
nikomakela: The real name of the children count field in study_group model is "group_size". That's a missleading name, so I wanted to rename it to the enrolment report model. I added this getter there for consistensy reasons.

nikomakela: so, group_size is not children + adults - it's just the children. Bad name from history.

charn (approved):
LGTM! :+1: :shipit:

PR #229, opened 2021-11-26
Occurrence upcoming filter renamed to enrollable and new real upcoming filter added

nikomakela:
PT-1310 PT-1371

### DIFF
--- a/occurrences/filters.py
+++ b/occurrences/filters.py
@@ -13,14 +13,30 @@ class OccurrenceFilter(django_filters.FilterSet):
     upcoming = django_filters.BooleanFilter(
         method="filter_by_upcoming", field_name="start_time"
     )
+    enrollable = django_filters.BooleanFilter(
+        method="filter_by_enrollable", field_name="start_time"
+    )
     date = django_filters.DateFilter(lookup_expr="date", field_name="start_time")
     time = django_filters.TimeFilter(method="filter_by_time", field_name="start_time")
 
     class Meta:
         model = Occurrence
-        fields = ["upcoming", "date", "time", "p_event", "cancelled"]
+        fields = [
+            "upcoming",
+            "enrollable",
+            "date",
+            "time",
+            "p_event",
+            "cancelled",
+        ]
 
     def filter_by_upcoming(self, qs, name, value):
+        if value:
+            # Only work with PostgreSQL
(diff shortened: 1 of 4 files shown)


occurrences/tests/test_api.py
```diff
@@ -2343,8 +2345,36 @@ def test_occurrences_filter_by_time(
...
+
+    # Past occurrences
+    OccurrenceFactory(
+        start_time=datetime(2020, 1, 5, 0, 0, 0, tzinfo=timezone.now().tzinfo),
+        p_event=p_event_1,
+    )
+
+    executed = api_client.execute(OCCURRENCES_QUERY, variables={"upcoming": True})
+    snapshot.assert_match(executed)
+    assert len(executed["data"]["occurrences"]["edges"]) == 3
```
tuomas777, line 2373: Personally in tests like this, I like to check that the returned instances are the ones we are expecting to make sure the tests are not passing falsely.

Also, isn't this returning all the occurrences (3), but one is in the past and should not be returned I assume?

occurrences/filters.py
```diff
@@ -13,14 +13,30 @@ class OccurrenceFilter(django_filters.FilterSet):
...
+            "enrollable",
+            "date",
+            "time",
+            "p_event",
+            "cancelled",
+        ]
 
     def filter_by_upcoming(self, qs, name, value):
+        if value:
+            # Only work with PostgreSQL
```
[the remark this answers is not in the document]
nikomakela: I have doubted it as well. :)

occurrences/tests/test_api.py
nikomakela: Yea, sorry that "past occurrence" comment was left there from a copy-paste. 👎 3 is the right result. There is a snapshot assertion that checks the right objects. Because this test does not have a database connection, I cannot test occurrence ids. I cannot compare them directly either, because the other one is an graphql document and another one is an ORM instance. The count test is there to make the test clearer and easier to understand, but it didn't work this time I guess :D 

If really wanted, I can remove the snapshot test or the count tes and compare dates or something, but I don't see that it would be good to test the both, the snapshot and the dates.

tuomas777: Blah, I managed to miss that there is also a snapshot in this test, my bad.

tuomas777 (approved):
LGTM :shipit:

PR #230, opened 2021-11-30
Added a Size2ArrayFieldBase serializer for "tuple fields"

nikomakela:
PT-1369

image

### DIFF
--- a/reports/serializers.py
+++ b/reports/serializers.py
@@ -1,8 +1,42 @@
+from typing import List
+
 from reports.models import EnrolmentReport
 from rest_framework import serializers
 
 
+class Size2ArrayField(serializers.Field):
+    def __init__(self, first_value_name: str = "id", second_value_name: str = "label"):
+        self.first_value_name = first_value_name
+        self.second_value_name = second_value_name
+        super().__init__()
+
+    def _array_to_dict(self, entry: List[str]):
+        if len(entry) != 2:
+            raise ValueError(
+                "Size2ArrayField field value should always "
+                "have a length of 2 (id, interpretation)."
+            )
+        return {self.first_value_name: entry[0], self.second_value_name: entry[1]}
+
+    def _dict_to_array(self, entry: dict):
+        return [entry[self.first_value_name], entry[self.second_value_name]]
+
+    def to_representation(self, value):
+        return [self._array_to_dict(entry) for entry in value] if value else None
+
+    def to_internal_value(self, data):
+        return [self._dict_to_array(entry) for entry in data] if data else None
+
+
 class EnrolmentReportSerializer(serializers.ModelSerializer):
+    study_group_study_levels = Size2ArrayField(
+        first_value_name="id", second_value_name="label"
+    )
+    occurrence_languages = Size2ArrayField(
+        first_value_name="id", second_value_name="name"
+    )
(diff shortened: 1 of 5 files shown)


reports/serializers.py
```diff
@@ -1,8 +1,51 @@
+from typing import List
+
 from reports.models import EnrolmentReport
 from rest_framework import serializers
 
 
+class Size2ArrayFieldBase(serializers.Field):
+    def __init__(self, first_value_name: str = "id", second_value_name: str = "label"):
```
tuomas777, line 8: Optimally this should allow `kwargs` to support `Field`'s [core arguments](https://www.django-rest-framework.org/api-guide/fields/#core-arguments).

reports/serializers.py
```diff
@@ -1,8 +1,51 @@
...
+
+
+class OccurrenceLanguagesField(Size2ArrayFieldBase):
+    def __init__(self):
+        super().__init__("id", "name")
+
+
+class EventKeywordsField(Size2ArrayFieldBase):
+    def __init__(self):
+        super().__init__("id", "name")
```
tuomas777, line 41: Not sure why these subclasses are needed, wouldn't it be easier to just use `Size2ArrayField` in serializers like `keywords = Size2ArrayField(second_value_name="name")`.

reports/tests/test_serializers.py
```diff
@@ -0,0 +1,46 @@
...
+
+def test_enrolment_report_to_representation():
+    report = EnrolmentReport(
+        occurrence_languages=[["fi", "Finnish"]],
+        study_group_study_levels=[["age_0_2", "age 0-2"]],
+        keywords=[["kultus:5", "Musiikki"]],
+    )
+    serializer = EnrolmentReportSerializer(instance=report)
+    assert serializer.data["occurrence_languages"][0]["id"] == "fi"
+    assert serializer.data["occurrence_languages"][0]["name"] == "Finnish"
```
tuomas777, line 13:
```suggestion
    assert serializer.data["occurrence_languages"][0] == {"id": "fi", "name": "Finnish"}
```
would be more succinct and also verify that there is no extra data in the object.

reports/serializers.py
```diff
@@ -1,8 +1,51 @@
...
+class Size2ArrayFieldBase(serializers.Field):
+    def __init__(self, first_value_name: str = "id", second_value_name: str = "label"):
+        self.first_value_name = first_value_name
+        self.second_value_name = second_value_name
+        super().__init__()
+
+    def _array_to_dict(self, entry: List[str]):
+        if len(entry) == 2:
+            return {self.first_value_name: entry[0], self.second_value_name: entry[1]}
+        return None
```
tuomas777, line 16: Do we really want to just silently ignore this error situation here? 🙂

reports/serializers.py
```diff
@@ -1,8 +1,42 @@
+from typing import List
+
 from reports.models import EnrolmentReport
 from rest_framework import serializers
 
 
+class Size2ArrayFieldBase(serializers.Field):
```
tuomas777, line 7: Kinda nitpick, but now that this is an actual field that is used, it would probably make sense to remove `Base` from the name.

tuomas777 (approved):
One more naming nitpick comment, other than that looks great to me 👍 🚀

PR #231, opened 2021-12-01
New page centric meta data for upcoming occurrences schema

nikomakela:
PT-1367

image

### DIFF
--- a/graphene_linked_events/schema.py
+++ b/graphene_linked_events/schema.py
@@ -1,4 +1,5 @@
 import json
+import math
 from types import SimpleNamespace
 
 from django.core.exceptions import PermissionDenied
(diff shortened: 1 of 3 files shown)


graphene_linked_events/schema.py
```diff
@@ -474,8 +498,8 @@ def resolve_events(parent, info, **kwargs):
     @staticmethod
     def resolve_upcoming_events(parent, info, **kwargs):
         """Return a list of upcoming events based on occurrence start time."""
-        page = kwargs.get("page")
-        page_size = kwargs.get("page_size")
+        page = kwargs.get("page", 1)
```
charn, line 501: There's the `default_value` defined, so I think this should be 1 if the argument is not given. :thinking: Or is there a scenario where the `default_value` does not work and we need to have this here?

charn (approved):
Left a minor comment, but LGTM! :+1: :shipit:

graphene_linked_events/schema.py
nikomakela: I can take them away. I added them because the method itself does not tell that a default value is set. Just by looking that method, it does not tell that there is a default value and I was afraid that it could be left as None.

PR #232, opened 2021-12-02
Added the Extra needs field to enrolment reports

nikomakela:
1. Added the Extra needs field to enrolment reports
2. Changed Extra needs -field to a TextField and fixed its verbose name
3. Removed the Unit id -fields required attribute in study group admin form

PT-1360

image

image


occurrences/admin.py
```diff
@@ -49,6 +49,11 @@ class StudyGroupAdmin(admin.ModelAdmin):
     list_filter = ["created_at", HasUnitIdStudyGroupListFilter]
     search_fields = ["unit_id", "unit_name", "person"]
 
+    def get_form(self, request, obj=None, **kwargs):
+        form = super().get_form(request, obj, **kwargs)
+        form.base_fields["unit_id"].required = False
```
charn, line 54: This looks like something that would be resolved with havin `blank=True` on the model.

nikomakela: But then we go back to our earlier discussion where we talked about whether or not the unit_id should ever be blank 😄 I would say, again, that this is a weakness in Django. The framework should not ever handle a nullable field as a blankable, because null and ´´really are not the same thing and when they are queried, they both needs a check. It's bad for data integrity - Really bad actually, because when a data report is done, you need to be aware of this.

nikomakela: okay, of course I could set the required_fields as a custom field, but I would still need to override the get_form -method.

nikomakela: Here are only the nullable fields:
```
nullable_fields = [field.name for field in StudyGroup._meta.get_fields() if field.concrete and field.null and not getattr(field, 'blank', None)]
```

PR #234, opened 2021-12-02
Enrolment summary uses only enrolments which occurrence is in future.

nikomakela:
PT-1364 - Enrolment summary emails fails sending because they couldn't fetch events from LinkedEvents. I found out  that summary tries to fetch all pending enrolments from history - to the beginning of time. LinkedEvents deletes events from past.

Also did some refactoring related to the notification sending: Moved the functions where I think they belong, since QuerySet should not send any emails.


nikomakela:
NOTE: The last commit contains the actual changes needed. Others are for refactoring.

occurrences/models.py
```diff
@@ -472,60 +471,34 @@ def save(self, *args, **kwargs):
...
+    def pending_enrolments_by_email(self, email: str):
+        return self.filter(
+            occurrence__p_event__contact_email=email, status=Enrolment.STATUS_PENDING,
+        )
 
-    def get_by_unique_id(self, unique_id):
-        compound_id = get_node_id_from_global_id(unique_id, "EnrolmentNode")
-        enrolment_id, ts = compound_id.split("_")
-        return self.get(id=enrolment_id, enrolment_time=ts)
+    def new_enrolments_by_email(self, email: str):
```
charn, line 498: or `approved_enrolments_by_email` ?

nikomakela: Absolutely! I'll change it

charn (approved):
LGTM! :+1: :shipit:

PR #236, opened 2021-12-03
Fix to manually approved study groups enrolments capacity validation

nikomakela:
PT-1392


nikomakela:
NOTE: Snapshot shows that there is still some issues

charn (approved):
LGTM! :+1: :shipit:

PR #239, opened 2021-12-08
PT-1289 | Search nearby events

charn:
This PR adds `nearbyPlaceId` and `nearbyDistance` arguments to the `events` query, which will allow searching for nearby events. PR also contains some refactoring for distance calculations after adding `geopy` to the requirements.


nikomakela (approved):
Looks good and tests proves that it works :) Good job! I had some issues with typing-extension on my mac (python 3.7.9), but the latest force push fixed them.

PR #243, opened 2021-12-29
Added translatable auto acceptance message field

nikomakela:
Added translatable auto acceptance message field to PalvelutarjotinEvent.

- Created a new proxy model TranslatedPalvelutarjotinEvent that offers a translatable auto acceptance message field for PalvelutarjotinEvent.
- Added the new translatable auto acceptance message field to the GraphQL schema.

PT-1400 PT-1433


graphene_linked_events/tests/snapshots/snap_test_api.py
```diff
@@ -739,6 +740,7 @@
                     "offers": [{"isFree": False}],
                     "pEvent": {
                         "autoAcceptance": True,
+                        "autoAcceptanceMessage": None,
```
nikomakela, line 743: - [ ] autoAcceptanceMessage should not be `None`

nikomakela: - [ ] IT should not be `{fi:null}` either :)

nikomakela: Finally ended up to use the same style as occurrences schema uses with study levels and venues. In my opinion it's not as good solution as it would be to use the translations in "map format", which is the way how the LinkedEvents returns the translations, but at least it is easier to maintain when more supported languages are added and all the same infromation is still available - it's just in a list format.

nikomakela:
I first implemented all this without Parler and the implementation was a lot easier! But then I thought that we will need to support the translations sooner or later, probably sooner, so it would make more sense to take care of translation data structure right from the beginning, because then we would avoid painful data migrations. It's up to UI whether it supports the translations or not, but the database and the backend API now does.

nikomakela:
- [ ] When UI does not support translations yet, I think the API should allow implementation with `autoAcceptanceMessage: string`. Currently it accepts only via `translations` list.

nikomakela:
The TranslatedPalvelutarjotinEvent proxy model is removed in this alternative approach proposed in PR #247

nikomakela:
No proxy model for PalvelutarjotinEvent translations anymore! 👍

PR #246, opened 2022-01-07
Enrolment reports divisions information given in more detailed format

nikomakela:
PT-1421

NOTE: Extends PT-1420 #245 .

image


reports/serializers.py
```diff
@@ -38,9 +38,113 @@ def to_internal_value(self, data):
...
+                **{
+                    self.translations[key]
+                    if key in self.translations
+                    else key: value.capitalize()
+                    for (key, value) in pairs
+                },
+            }
+        return None
+
+    def to_internal_value(self, data: dict) -> list:
```
charn, line 97: I guess we don't actually need to support deserialization since the reporting API is not taking any input, but since this is already implemented and tested there's no harm in having it. :+1:

charn (approved):
LGTM! :+1: :shipit:

reports/serializers.py
```diff
@@ -38,9 +38,113 @@ def to_internal_value(self, data):
...
+            dict: the original OCD ids list in the key "ocd-ids" and
+            all the key-value -pairs of OCD id path, e.g.
+            ```
+            "ocd_ids": [
+                "ocd-division/country:fi/kunta:helsinki/osa-alue:keski-pasila",
+                "ocd-division/country:fi/kunta:helsinki/kaupunginosa:pasila",
+                "ocd-division/country:fi/kunta:helsinki/peruspiiri:pasila",
+                "ocd-division/country:fi/kunta:helsinki",
+            ],
+            "country": "Fi",
```
tuomas777, line 76: Uppercase `FI` would be the proper ISO country code.

tuomas777 (approved):
Added a couple of minor comments, LGTM 👍

PR #250, opened 2022-01-18
Fixes and improvements to LinkedEvents API error handlings

nikomakela:
PT-1461. Fixed a null pointer exception handling related to empty location. Also raising the errors from LinkedEvents queries done by schema.


occurrences/management/commands/send_enrolment_summary.py
```diff
@@ -1,8 +1,7 @@
+from django.conf import settings
 from django.core.management.base import BaseCommand
 from occurrences.models import Enrolment
 
-from palvelutarjotin import settings
```
charn, line 4: I quickly checked that same import is done in a few other places as well. Maybe those should also be fixed/changed like this?

graphene_linked_events/schema.py
```diff
@@ -810,10 +819,10 @@ def mutate(root, info, **kwargs):
     def _create_p_event(event_obj, organisation, p_event_data) -> PalvelutarjotinEvent:
         p_event_data["linked_event_id"] = event_obj.id
         p_event_data["organisation_id"] = organisation.id
-        p_event_data["organisation_id"] = organisation.id
-        translations = p_event_data.pop("translations")
+        translations = p_event_data.pop("translations", None)
```
nikomakela, line 822: The event will be created in LinkedEvents before the PalvelutarjotinEvent is created. When the PalvelutarjotinEvent creation was a failure because of missing translations, the event was created only in the LE side.

I added a fix and a test for missing translations, but in new version of the UI, the translations should always be populated, so this should have never risen up. The logs shows that Hurjaruuth user ha an old version of the event creation wizard / form.

Validation before the event would be too much work and just make the mess of code. Lets try to manage with the little fix done.

charn (approved):
LGTM! :+1: :shipit:

nikomakela:
@charn one more addition to this PR. If the PalvelutarjotinEvent fails in creation, we could delete the Event from LinkedEvents side to prevent gettingempty pages like we did get with this production support ticket https://helsinkisolutionoffice.atlassian.net/browse/PT-1458

graphene_linked_events/schema.py
```diff
@@ -800,7 +812,19 @@ def mutate(root, info, **kwargs):
         event_obj = json2obj(format_response(result))
         if result.status_code == 201:
             # Create palvelutarjotin event if event created successful
-            AddEventMutation._create_p_event(event_obj, organisation, p_event_data)
+            try:
+                AddEventMutation._create_p_event(event_obj, organisation, p_event_data)
+            except Exception as e:
+                logger.error(
```
charn, line 818: This could also be `logger.exception` which I think should automatically put the exception information into to the log.

charn (approved):
LGTM! :+1: :shipit:

PR #258, opened 2022-04-21
PT-1506 | Event list requests to Linked events always include data source parameter

charn:
Data source is required for the LinkedEvents production API to return any palvelutarjotin related events.


nikomakela (approved):
Good!

PR #268, opened 2022-07-11
Memory leak

terovirtanen:
uwsgi-module memory limit [PT-1184](https://helsinkisolutionoffice.atlassian.net/browse/PT-1184)
horizontal pod scaler [DEVOPS-392](https://helsinkisolutionoffice.atlassian.net/browse/DEVOPS-392)
load test for palvelutarjotin-api.test.kuva.hel.ninja [DEVOPS-428](https://helsinkisolutionoffice.atlassian.net/browse/DEVOPS-428)

### DIFF
--- a/k6LoadTests.js
+++ b/k6LoadTests.js
@@ -14,7 +14,7 @@ export const options = {
 
 export default () => {
   const url = 'https://palvelutarjotin-api.test.kuva.hel.ninja/graphql';
-  const data = 'query=query Places {places(showAllPlaces: true){data{id}}}';
+  const data = 'query=query Organisations {organisations {edges {node {id } } } }';
   const res = http.post(url, data, {
     headers: { 'Content-Type': 'application/x-www-form-urlencoded' },
   });


k6LoadTests.js
```diff
@@ -0,0 +1,24 @@
...
+  //  vus: 1,
+  thresholds: {
+    //avg is around ?800ms? on https://palvelutarjotin-api.test.kuva.hel.ninja
+    http_req_duration: ['p(95)<5000'],
+  },
+};
+
+export default () => {
+  const url = 'https://palvelutarjotin-api.test.kuva.hel.ninja/graphql';
+  const data = 'query=query Places {places(showAllPlaces: true){data{id}}}';
```
nikomakela, line 17: If the purpose of this is to only "make some load", I would prefer some another query, e.g. organisations, because the place-query actually fetches the data from the LinkedEvents API.

```suggestion
  const data = 'query=query Organisations {organisations {edges {node {id } } } }';
```

nikomakela: It's another thing if we really want to test with API dependencies!

k6LoadTests.js
```diff
@@ -0,0 +1,24 @@
...
+  duration: '10m',
+  vus: 20,
+  //  vus: 1,
+  thresholds: {
+    //avg is around ?800ms? on https://palvelutarjotin-api.test.kuva.hel.ninja
+    http_req_duration: ['p(95)<5000'],
+  },
+};
+
+export default () => {
```
nikomakela, line 15: Would it make any sense to add this K6 to the docker-compose configuration? Installation instructions: https://k6.io/docs/getting-started/installation/#docker.

.prod/uwsgi.ini
```diff
@@ -6,6 +6,13 @@ static-map = /static=/var/static
 static-map = /media=/app/var/media
 uid = appuser
 gid = appuser
+
+# Reload workers regularly to keep memory fresh
+# and ease potential memory leaks
+max-requests = 1000                  # Restart workers after this many requests
```
nikomakela, line 12: What is our default for this? Or the current value? I found a post from the StackOverflow that prefers these (https://stackoverflow.com/a/44641109/784642):

```
Haharakiri = 20 # respawn processes taking more than 20 seconds
limit-as = 256 # limit the project to 256 MB
max-requests = 5000 # respawn processes after serving 5000 requests
daemonize = /var/log/uwsgi/yourproject.log # background the process & log
```

nikomakela (approved):

PR #269, opened 2022-07-15
Set workflows concurrency

terovirtanen:
Set workflow concurrency [DEVOPS-429](https://helsinkisolutionoffice.atlassian.net/browse/DEVOPS-429)
Load test is run after staging deployment is ready

### DIFF
--- a/.github/workflows/review.yml
+++ b/.github/workflows/review.yml
@@ -38,6 +38,9 @@ jobs:
     runs-on: ubuntu-latest
     needs: build
     name: Review
+    # allow only one deployment at a time
+    concurrency: ${{ github.workflow }}-${{ github.head_ref }}
+
     steps:
       - uses: actions/checkout@v2
       - uses: andersinno/kolga-setup-action@v2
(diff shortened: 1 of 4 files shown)


.github/workflows/review.yml
```diff
@@ -38,6 +38,9 @@ jobs:
     runs-on: ubuntu-latest
     needs: build
     name: Review
+    # allow only one deployment at a time
+    concurrency: ${{ github.workflow }}-${{ github.head_ref }}
```
nikomakela, line 42: what happens if I merge 2 pull requests? Will them go in queue or will one of them get dismissed?

nikomakela (commented):
Looks good, but I need a confirmation to question about queuing

.github/workflows/review.yml
terovirtanen: Review workflow
- pull requests are run separately and parallel like now. 
- 2 (or more) commits to same pull request at the same time:
  - build -steps are run parallel as now
  - deploy -steps to Kubernetes are run on queue

Staging workflow
- merge 2 (or more) pull request at the same time to master triggers 2 actions where:
  - build -steps are run parallel as now
  - deploy -steps to Kubernetes are run on queue

Production workflow
- all release actions are run on queue (this is more theoretical case)

nikomakela (approved):

PR #272, opened 2022-09-06
Make SMS service on-feature configurable

nikomakela:
PT-1538. Because of the ongoing situation in production server, where the enrolments sends SMS-messages wrongly, the first step is to make it possible to switch of the SMS-sending feature as a whole.

Also needed to install new version of pre-commit tools, including Black, (Isort and Flake8), or otherwise I could not commit with verification hooks on, because of an issue in the Black. See, https://exerror.com/importerror-cannot-import-name-_unicodefun-from-click/ or https://stackoverflow.com/questions/71673404/importerror-cannot-import-name-unicodefun-from-click


nikomakela:
CI is failing because this needs some black-formatting that is done in another pull request: #274 .

nikomakela:
If the settings are wanted to be enabled in the production and they are wanted to be enabled with permanent configuration, there is PR #273

nikomakela:
The actual solution should be found from the PR #275

karisal-anders:
The isort/black/flake8 versions are now a bit out-of-sync:

|package|.pre-commit-config.yaml|requirements-dev.txt|
|---|---|---|
|isort| v5.10.1 | 4.3.21 |
|black|22.8.0|22.8.0|
|flake8|3.9.2 | 3.7.8 |

Updating all of those to exactly match each other would be ideal IMHO to remove any possible discrepancies between development environment's source code formatting and the precommit hook. Or are there some assurances somewhere that e.g. inside major versions some aspects don't change for isort/black/flake8?

karisal-anders (approved):

tuomas777:
Not a huge thing, but I think now other SMS tests will fail if SMS sending is disabled in the env where the tests are run.

nikomakela:
> Not a huge thing, but I think now other SMS tests will fail if SMS sending is disabled in the env where the tests are run.

So I guess the only thing that could help with this is to initialise a new file for pytest-configuration (https://pytest-django.readthedocs.io/en/latest/configuring_django.html#pytest-ini-settings). But. I would  say that this is not the only setting that would fail in such cases. I guess I create a new ticket of that, but won't fight with that now

karisal-anders:
> > Not a huge thing, but I think now other SMS tests will fail if SMS sending is disabled in the env where the tests are run.
> 
> So I guess the only thing that could help with this is to initialise a new file for pytest-configuration (https://pytest-django.readthedocs.io/en/latest/configuring_django.html#pytest-ini-settings). But. I would say that this is not the only setting that would fail in such cases. I guess I create a new ticket of that, but won't fight with that now

Another possibility would be to use `django.test.override_settings(NOTIFICATION_SERVICE_SMS_ENABLED=True)` where needed to document that these test cases rely on this setting to be true. This may get a bit verbose though if needed to be put into a lot of tests so if it's needed in a lot of places having that ini settings sounds like a better idea.

tuomas777:
> > > Not a huge thing, but I think now other SMS tests will fail if SMS sending is disabled in the env where the tests are run.
> > 
> > 
> > So I guess the only thing that could help with this is to initialise a new file for pytest-configuration (https://pytest-django.readthedocs.io/en/latest/configuring_django.html#pytest-ini-settings). But. I would say that this is not the only setting that would fail in such cases. I guess I create a new ticket of that, but won't fight with that now
> 
> Another possibility would be to use `django.test.override_settings(NOTIFICATION_SERVICE_SMS_ENABLED=True)` where needed to document that these test cases rely on this setting to be true. This may get a bit verbose though if needed to be put into a lot of tests so if it's needed in a lot of places having that ini settings sounds like a better idea.

I can be wrong, but by quickly looking at the tests, there seems to be only [one other test](https://github.com/City-of-Helsinki/palvelutarjotin/blob/c72b6eba71db0f4666e8645d078cc495d3786477/occurrences/tests/test_notifications.py#L86) that uses SMSs. If that is the case, the best solution for now is probably just overriding settings in that?

Regarding more or less global test settings, I've also successfully used Pytest fixtures with `autouse=True` for that.

nikomakela:
> > > > Not a huge thing, but I think now other SMS tests will fail if SMS sending is disabled in the env where the tests are run.
> > > 
> > > 
> > > So I guess the only thing that could help with this is to initialise a new file for pytest-configuration (https://pytest-django.readthedocs.io/en/latest/configuring_django.html#pytest-ini-settings). But. I would say that this is not the only setting that would fail in such cases. I guess I create a new ticket of that, but won't fight with that now
> > 
> > 
> > Another possibility would be to use `django.test.override_settings(NOTIFICATION_SERVICE_SMS_ENABLED=True)` where needed to document that these test cases rely on this setting to be true. This may get a bit verbose though if needed to be put into a lot of tests so if it's needed in a lot of places having that ini settings sounds like a better idea.
> 
> I can be wrong, but by quickly looking at the tests, there seems to be only [one other test](https://github.com/City-of-Helsinki/palvelutarjotin/blob/c72b6eba71db0f4666e8645d078cc495d3786477/occurrences/tests/test_notifications.py#L86) that uses SMSs. If that is the case, the best solution for now is probably just overriding settings in that?
> 
> Regarding more or less global test settings, I've also successfully used Pytest fixtures with `autouse=True` for that.

Fixed with fixture

tuomas777 (approved):
LGTM 👍 🐷

PR #275, opened 2022-09-06
Send approval notifications after auto accepted enrolments are handled

nikomakela:
PT-1538. Preventing the SMS notification to be sent when enrolling to multiple occurrences at once and an error occurs after the first enrolment is handled. The transaction should and has rolled back, but the issue was in the SMS sending, which is sent via notification-service and the SMS notification was not rolled back.

This PR is done on the code formatting done in #274

### DIFF
--- a/occurrences/schema.py
+++ b/occurrences/schema.py
@@ -10,7 +10,7 @@ from graphene.utils.str_converters import to_snake_case
 from graphene_django import DjangoConnectionField, DjangoObjectType
 from graphene_django.filter import DjangoFilterConnectionField
 from graphql_jwt.decorators import staff_member_required
-from typing import List, Union
+from typing import List, Optional, Tuple, Union
 
 from common.utils import (
     get_editable_obj_from_global_id,
@@ -747,6 +747,7 @@ class EnrolOccurrenceMutation(graphene.relay.ClientIDMutation):
         study_group = _create_study_group(kwargs.pop("study_group"))
         contact_person_data = kwargs.pop("person", None)
         enrolments = []
+        notifiable_enrolments: List[Tuple[Enrolment, Optional[str]]] = []
         for occurrence_gid in occurrence_gids:
             occurrence_id = get_node_id_from_global_id(occurrence_gid, "OccurrenceNode")
             try:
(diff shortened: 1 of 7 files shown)


nikomakela:
Both, the batch approval and the normal multi enrolment should be working now, but I just cannot figure out why the solution to them is different. Why isn't the batch approval sending the SMS-messages, but the multi enrolment to an occurrence with auto acceptance did?

occurrences/schema.py
```diff
@@ -766,12 +767,18 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
...
                 )
-                enrolment.approve(custom_message=custom_message)
+                enrolment.approve(send_notification=False)
+                notifiable_enrolments.append((enrolment, custom_message))
             enrolments.append(enrolment)
 
+        for enrolment, custom_message in notifiable_enrolments:
+            enrolment.send_approve_notification(custom_message=custom_message)
+
         return EnrolOccurrenceMutation(enrolments=enrolments)
```
tuomas777, line 782: Does this work now when `auto_acceptance` is `False`? Based on the code it seems that a message gets sent by a signal handler [here](https://github.com/City-of-Helsinki/palvelutarjotin/blob/7608a837b3aec4a6a530fc097e567b45917a2f0f/occurrences/signals.py#L18-L32), which seems to have the same original problem?

nikomakela: The `approve` -method sent the notifications earlier: https://github.com/City-of-Helsinki/palvelutarjotin/pull/275/files/1a15b669c9801470782c6dbd19d978c1f13453ef#diff-6122150bb78e461eabd410f773e54d0014f1bbd9ecc32e368f59631bd7373892L631. Now I just splitted that part of the code to a new model-function that can be called separately.

The signal seems to send the messages only in cases of when the auto acceptance is off. The new code in schema, won't have anything in `notifiable_enrolments` when the auto-acceptance is off, so it won't send in those cases and it never has.

tuomas777: > the issue is only with enrolments that are for multiple occurrences all at once. In such cases, the auto acceptance cannot be on

If I understood correctly, I think in that sentence the last word should be `off`? 🙂 

But yeah, if that is the case, then I guess it will work. Seems pretty complicated indeed, and I would say that the signal thing should be refactored away to avoid future problems 🙂 Not needed in this PR ofc if/when this is in hurry.

nikomakela: yep, edited the comment :)

nikomakela:
some new tests were removed in the rebase process accidentally, so I needed to rebase and force-push a bit.

occurrences/schema.py
```diff
@@ -747,6 +747,7 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
         study_group = _create_study_group(kwargs.pop("study_group"))
         contact_person_data = kwargs.pop("person", None)
         enrolments = []
+        notifiable_enrolments: List[Tuple(Enrolment, Optional[str])] = []
```
karisal-anders, line 750: Did you mean for the type to be `List[Tuple[Enrolment, Optional[str]]]`?

karisal-anders (approved):
Looks ok to merge for me

PR #276, opened 2022-09-09
Disallow enrolling with an empty group

nikomakela:
PT-1542 PT-1548

Also refactored some old tests related to enrolments to clean the code.

### DIFF
--- a/occurrences/tests/test_api.py
+++ b/occurrences/tests/test_api.py
@@ -1442,10 +1442,12 @@ def test_enrol_invalid_group_size(
 # The auto_acceptance calls Enrolment.approve,
 # so it needs to be tested with both the boolean values.
 @pytest.mark.parametrize("auto_acceptance", [True, False])
+@pytest.mark.parametrize("is_multi_enrolment", [True, False])
 @patch("occurrences.services.notification_service.send_sms")
 def test_enrol_full_people_count_seat_type_occurrence(
     mock_send_sms,
     auto_acceptance,
+    is_multi_enrolment,
     api_client,
     mock_update_event_data,
     mock_get_event_data,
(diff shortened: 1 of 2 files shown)


occurrences/tests/test_api.py
```diff
@@ -1463,68 +1468,8 @@ def test_enrol_full_people_count_seat_type_occurrence(
...
-    study_group_15 = StudyGroupFactory(group_size=15)
-    study_group_20 = StudyGroupFactory(group_size=20)
-    # Current date froze on 2020-01-04:
-    p_event_1 = PalvelutarjotinEventFactory(
-        enrolment_start=datetime(2020, 1, 3, 0, 0, 0, tzinfo=timezone.now().tzinfo),
-        enrolment_end_days=2,
-        auto_acceptance=auto_acceptance,
-    )
+    # if is_multi_enrolment is True, multiple occurrences are needed
     valid_occurrence = OccurrenceFactory(
```
karisal-anders, line 1510: If you want to do purely refactoring here (i.e. not changing functionality) then adding a
```python
if is_multi_enrolment:
    valid_occurence = ...
```
would make the code work more like before (i.e. not creating another Occurence when is_multi_enrolment is False).

occurrences/tests/test_api.py
```diff
@@ -1533,6 +1478,7 @@ def test_multi_enrol_full_people_count_seat_type_occurrence(
         amount_of_seats=100,  # enough for 2 enrolments
     )
 
+    # Ibvalid occurrence which does not support such a large study group
```
karisal-anders, line 1518: Typo: Ibvalid→Invalid

occurrences/tests/test_api.py
```diff
@@ -1439,6 +1439,43 @@ def test_enrol_invalid_group_size(
...
+                "groupName": empty_study_group.group_name,
+                "studyLevels": [
+                    sl.upper() for sl in empty_study_group.study_levels.all()
+                ],
+                "amountOfAdult": empty_study_group.amount_of_adult,
+            },
+        }
+    }
+    executed = api_client.execute(ENROL_OCCURRENCE_MUTATION, variables=variables)
+    assert_match_error_code(executed, INVALID_STUDY_GROUP_SIZE_ERROR)
```
karisal-anders, line 1476: Not necessary to do as a part of this pull request but combining the `variables` generating functionality from these kind of tests into a function of its own or somewhere to be reused would reduce code size.

occurrences/tests/test_api.py
```diff
@@ -1442,10 +1442,15 @@ def test_enrol_invalid_group_size(
 # The auto_acceptance calls Enrolment.approve,
 # so it needs to be tested with both the boolean values.
 @pytest.mark.parametrize("auto_acceptance", [True, False])
+@pytest.mark.parametrize(
+    "is_multi_enrolment",
+    [True, False],
```
karisal-anders, line 1484: Not necessary to do (i.e. you can leave this as is if you want to): Removing the trailing comma and rerunning black would fit this on a single line.

karisal-anders (commented):
Left some comments, please take a look at them.

occurrences/tests/test_api.py
nikomakela: had it like that, but then it would have been Outbound. Did a small refactor here now.

occurrences/tests/test_api.py
nikomakela: Yea, would be a huge change to change them all now. And it's actually hard to follow what happens in the test if it's built fully from utils. But I'm also quite sure that there could be an easily understandable utility function for this.

occurrences/tests/test_api.py
karisal-anders: > had it like that, but then it would have been Outbound. Did a small refactor here now.

What do you mean by Outbound? Wanted to be sure and checked it locally that the test passes with the added if clause.

karisal-anders (approved):
Looks ok for me to merge!

PR #281, opened 2022-12-28
PT-1616 | Support for unauthenticated personal data deletion

tuomas777:
Implemented changes that allow unauthenticated (don't have a related `User` object) `Person` objects to be deleted even when they are related to an `Enrolment` or a `StudyGroup`. The related `Enrolment` and the `StudyGroup` objects won't be affected by the deletion.

The need for this comes from GDPR where users can request their personal data to be deleted, and too old personal data should be deleted automatically.


reports/templates/reports/admin/event_enrolments.html
```diff
@@ -44,8 +44,7 @@
             </td>
             <td>{{enrolment.study_group.group_size}}</td>
             <td>{{enrolment.study_group.amount_of_adult}}</td>
-            <td>{{enrolment.study_group.extra_needs}}</td>
```
nikomakela, line 47: should the extra needs column be removed?? I guess not, but if it should, the title should be removed too.

nikomakela (changes requested):
Nice job! 🥇  But 1 line too many removed!! 😱

reports/templates/reports/admin/event_enrolments.html
tuomas777: Well spotted sir! Fixed.

nikomakela (approved):

PR #282, opened 2022-12-30
PT-1617 | Implement Person retention period exceeded admin UI filter

tuomas777:


palvelutarjotin/settings.py
```diff
@@ -337,6 +338,8 @@
 
 MAX_UPLOAD_SIZE = 2621440  # 2MB
 
+PERSONAL_DATA_RETENTION_PERIOD_MONTHS = env("PERSONAL_DATA_RETENTION_PERIOD_MONTHS")
```
nikomakela, line 341: it would be a good practise to add these to `.env.example` -file and to `docker-compose.env.yaml.example`, but it seems like those does not include earlier params either. Also the CI configuration is not able to read this at the moment. But, I think this is fine like this, because I don't think we really change the value so often.

organisations/tests/test_models.py
```diff
@@ -127,3 +130,68 @@ def test_user_str():
...
+
+
+@pytest.mark.django_db
+def test_too_old_personal_data():
+    too_old = timezone.now() - relativedelta(months=24, days=1)
+    still_valid = timezone.now() - relativedelta(months=24, days=-1)
+
+    # ### EXPECTED PERSON OBJECTS IN RESULTS
+
+    expected_people = PersonFactory.create_batch(5, user=None)
```
nikomakela, line 142: trivial, but it would be better to "name" the persons here where the variable is defined, instead of using the indexes.

```suggestion
    old_enrolment_person, old_study_group_person, multi_enrolment_person, shared_all_old_person  = PersonFactory.create_batch(5, user=None)
```

organisations/tests/test_models.py
```diff
@@ -127,3 +130,68 @@ def test_user_str():
...
+    EnrolmentFactory(person=expected_people[0], occurrence__end_time=too_old)
+
+    # person with a too old study group
+    EnrolmentFactory(person=expected_people[1], occurrence__end_time=too_old)
+
+    # person with a too old enrolment and a too old study group
+    EnrolmentFactory(person=expected_people[2], occurrence__end_time=too_old)
+    EnrolmentFactory(
+        study_group__person=expected_people[2], occurrence__end_time=too_old
+    )
```
nikomakela, line 154: Is this even a possible scenario? Isn't it a whole new person then? Or is it possible when it's a multi-occurrence enrolment?

organisations/models.py
```diff
@@ -23,6 +25,30 @@ def user_can_view(self, user):
...
+            max_studygroup_end_time__lt=earliest_valid_timestamp
+        )
+
+        return (
+            # At least for now returns only unauthenticated Person's
+            self.filter(user=None)
+            .annotate(
+                max_enrolment_end_time=Max("enrolment__occurrence__end_time"),
+                max_studygroup_end_time=Max(
+                    "studygroup__enrolments__occurrence__end_time"
```
nikomakela, line 46: another one is in a plural form? "enrolments"?

nikomakela (approved):
Some notes, but approved

palvelutarjotin/settings.py
tuomas777: Yeah I agree. The idea was that most probably this value doesn't need to be changed ever, so it could be hard-coded, but because it needs practically zero effort to make it also env configurable, added that as well 🙂

organisations/tests/test_models.py
tuomas777: Good idea, changed.

organisations/tests/test_models.py
tuomas777: Probably not in practice 🙂. Removed.

organisations/models.py
tuomas777: The first one comes from `Enrolment` -> `Person` relation, where `related_name` is not set, so it defaults to the model's name. The second one comes from `Enrolment` -> `StudyGroup` relation, where `related_name` is set to `enrolments`.

PR #283, opened 2022-12-30
Authenticated personal data deletion support

tuomas777:
Implemented changes related to allowing authenticated users to be deleted properly via the admin UI:

- Implemented a method for deleting a `PalvelutarjotinEvent`'s contact info, so in practice fields `contact_person`, `contact_email` and `contact_phone_number`. There is also a new datetime field `contact_info_deleted_at` which is populated when the deletion takes place. **Note**: the deletion method needs to be executed explicitly, just emptying the fields will not be considered as a contact info deletion in this context.
- Added a management command that deletes contact info from too old events. The idea is that this command will be run periodically by cron.
- Added new section "enrollee person data" to the admin UI. The contents of that are technically `Person` objects that don't have a related `User` object.

  The motivation is to clearly separate unauthenticated personal data viewing and deletion from authenticated personal data. Previously both were handled via "person" section, which was pretty confusing and error-prone.
  
  Also from now on, the idea is that authenticated users should be deleted by deleting the `User` object, which will then remove correctly related `Person` and handle the deletion logic related to that.

Closes PT-1619 & PT-1621


organisations/models.py
```diff
@@ -56,10 +57,20 @@ def delete(self):
         return super().delete()
 
 
+class UserQueryset(models.QuerySet):
```
nikomakela, line 60: There is now something that affects to the admin login. The admin-login with this PR code is not currently possible. If I use the production version't commit to login and then change to this PR branch after successful login, all seems to work OK with the logged in user.

image

nikomakela (changes requested):
There is a login issue still! The rest of the changes looks good and seems to work OK. 

NOTE: The `OccurrenceAdmin`'s `contact_persons` field was a bit out of this scope, but it's currently quite useless 
image

organisations/models.py
nikomakela: @tuomas777 suggested a fix like this and I tested that it seems to work:
```
from django.contrib.auth.models import UserManager
class User(AbstractUser):
    objects = UserManager.from_queryset(UserQueryset)()
```

tuomas777:
> There is a login issue still! The rest of the changes looks good and seems to work OK.
> 
> NOTE: The `OccurrenceAdmin`'s `contact_persons` field was a bit out of this scope, but it's currently quite useless image

Actually, there is apparently no use for those Occurrence-related contact persons, and that relation should be removed altogether.

organisations/models.py
tuomas777: Added that fix to the code.

nikomakela (approved):

PR #288, opened 2023-01-16
Improvements to the enrolment reports LinkedEvents-data hydration

nikomakela:
PT-1525 PT-1527.
Add a has publisher filter to the enrolment reports admin page, because with that, it's easy to find the erroneous enrolment report entries and update them (using the admin actions).

Instead of throwing errors, just log them with a logger, so that more data is fetched. Instead of giving up when a first error occurs, we should try to fetch the rest of the info that is given. Also, an improvement to the logging was added, so that it's clearer where the issue is. It should also be noted that not all the places have coordinates, e.g. `helsinki:internet` for internet events, so we should not throw an KeyError in these cases, because it's a valid case.

### DIFF
--- a/reports/services.py
+++ b/reports/services.py
@@ -75,25 +75,18 @@ def get_place_json_from_linkedevents(place_id: str):
 
 
 def resolve_place_divisions(place_json) -> Optional[list]:
-    if not place_json:
-        return None
+    # When fetching the event, the division field is under location
+    # NOTE: Some places, e.g helsinki:internet, might not have divisions
     try:
-        # When fetching the event, the division field is under location
-        if "location" in place_json:
-            place_json = place_json["location"]
-        divisions = place_json["divisions"]
-        return [d["ocd_id"] for d in divisions]
-    except KeyError:
+        return [d["ocd_id"] for d in place_json["location"]["divisions"]]
+    except (IndexError, KeyError, TypeError):
         return None
 
 
 def resolve_place_coordinates(place_json) -> Optional[list]:
-    if not place_json:
-        return None
+    # When fetching the event, the position field is under location
+    # NOTE: Some places, e.g helsinki:internet, might not have coordinates
     try:
-        # When fetching the event, the position field is under location
-        if "location" in place_json:
-            place_json = place_json["location"]
-        return place_json["position"]["coordinates"]
-    except KeyError:
+        return place_json["location"]["position"]["coordinates"]
+    except (IndexError, KeyError, TypeError):
         return None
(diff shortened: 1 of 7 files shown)


reports/services.py
```diff
@@ -79,10 +79,15 @@ def resolve_place_divisions(place_json) -> Optional[list]:
...
+        if (
+            place_json
+            and "location" in place_json
+            and "divisions" in place_json["location"]
+        ):
+            divisions = place_json["location"]["divisions"]
+            return [d["ocd_id"] for d in divisions]
+        return None
     except KeyError:
         return None
```
karisal-anders, line 92: The `if place_json` in the try clause is redundant with the `if not place_json` clause before the try clause, only one of them is needed.

One quite idiomatic Python (i.e. ["Easier to ask for forgiveness than permission"](https://docs.python.org/3/glossary.html#term-EAFP)) refactoring of the function:
```python
def resolve_place_divisions(place_json) -> Optional[list]:
    # When fetching the event, the division field is under location
    # NOTE: Some places, e.g helsinki:internet, might not have divisions
    try:
        return [d["ocd_id"] for d in place_json["location"]["divisions"]]
    except (IndexError, KeyError, TypeError):
        return None
```
Although I gotta say coming from a C++ background that looks a bit dangerous to me but AFAIK that's the more idiomatic Python way to do it. What's your take on this?

reports/services.py
```diff
@@ -92,8 +97,11 @@ def resolve_place_coordinates(place_json) -> Optional[list]:
...
-            place_json = place_json["location"]
-        return place_json["position"]["coordinates"]
+        # NOTE: Some places, e.g helsinki:internet, might not have coordinates
+        if place_json:
+            if "location" in place_json:
+                place_json = place_json["location"]
+            return place_json["position"]["coordinates"]
+        return None
     except KeyError:
         return None
```
karisal-anders, line 107: The `if place_json` in the try clause is redundant with the `if not place_json` clause before the try clause, only one of them is needed.

One quite idiomatic Python (i.e. ["Easier to ask for forgiveness than permission"](https://docs.python.org/3/glossary.html#term-EAFP)) refactoring of the function:
```python
def resolve_place_coordinates(place_json) -> Optional[list]:
    # When fetching the event, the position field is under location
    # NOTE: Some places, e.g helsinki:internet, might not have coordinates
    try:
        return place_json["location"]["position"]["coordinates"]
    except (IndexError, KeyError, TypeError):
        return None
```
Although I gotta say coming from a C++ background that looks a bit dangerous to me but AFAIK that's the more idiomatic Python way to do it. What's your take on this?

karisal-anders (approved):
Overall looks good, left a few comments about redundancy and possible refactoring. As neither of those is critical I'll approve this already and you can decide whether to still improve the code a bit or not.

reports/services.py
nikomakela: Let's try that

PR #291, opened 2023-01-17
Save user's last activity if the configured interval has passed

nikomakela:
PT-1620.
Write the authentication timestamp to user's last_login -field on every 60 minutes.

### DIFF
--- a/palvelutarjotin/oidc.py
+++ b/palvelutarjotin/oidc.py
@@ -1,3 +1,6 @@
+from django.conf import settings
+from django.contrib.auth.models import update_last_login
+from django.utils import timezone
 from helusers.oidc import ApiTokenAuthentication
 
 
@@ -19,7 +22,31 @@ class FixTunnistamoMixin:
         super().validate_claims(id_token)
 
 
-class GraphQLApiTokenAuthentication(FixTunnistamoMixin, ApiTokenAuthentication):
+class FixLastLoginMixin:
+    def update_last_login_after_interval(self, user):
+        """
+        Update the user object last_login field if enough time has passed.
+        The setting UPDATE_LAST_LOGIN["UPDATE_INTERVAL_MINUTES"] determines
+        whether or not enough time has passed for update.
+        We don't want to update on every request,
+        because then there would be too much writes when there are lot of users.
+        """
+        minutes_since_last_update = None
+        if user.last_login:
+            minutes_since_last_update = (
+                timezone.now() - user.last_login
+            ).total_seconds() / 60
+        if (
+            minutes_since_last_update is None
+            or minutes_since_last_update
+            > settings.UPDATE_LAST_LOGIN["UPDATE_INTERVAL_MINUTES"]
+        ):
+            update_last_login(sender=None, user=user)
+
+
+class GraphQLApiTokenAuthentication(
+    FixTunnistamoMixin, FixLastLoginMixin, ApiTokenAuthentication
+):
     """
     Custom wrapper for the helusers.oidc.ApiTokenAuthentication backend.
     Needed to make it work with graphql_jwt.middleware.JSONWebTokenMiddleware,
@@ -34,6 +61,10 @@ class GraphQLApiTokenAuthentication(FixTunnistamoMixin, ApiTokenAuthentication):
         if not user_auth_tuple:
             return None
         user, auth = user_auth_tuple
+
+        if settings.UPDATE_LAST_LOGIN and settings.UPDATE_LAST_LOGIN["ENABLED"]:
+            self.update_last_login_after_interval(user)
+
         return user
 
 
--- a/palvelutarjotin/settings.py
+++ b/palvelutarjotin/settings.py
@@ -75,6 +75,8 @@ env = environ.Env(
     ENABLE_SUMMARY_REPORT=(bool, False),
     VERIFICATION_TOKEN_VALID_MINUTES=(int, 15),
     PERSONAL_DATA_RETENTION_PERIOD_MONTHS=(int, 24),
+    UPDATE_LAST_LOGIN_ENABLED=(bool, True),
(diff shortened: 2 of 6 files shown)


palvelutarjotin/settings.py
```diff
@@ -353,3 +356,8 @@
 NOTIFICATIONS_IMPORTER = (
     "notification_importers.notification_importer.NotificationFileImporter"
 )
+
+UPDATE_LAST_LOGIN = {
+    "ENABLED": env.bool("UPDATE_LAST_LOGIN"),
+    "UPDATE_INTERVAL": env.int("UPDATE_LAST_LOGIN_INTERVAL_MINUTES"),
+}
```
karisal-anders, line 363: The environment variable names UPDATE_LAST_LOGIN and UPDATE_LAST_LOGIN_INTERVAL_MINUTES seem transparent to me so why not just use them with direct mapping to settings? i.e.:
```python
UPDATE_LAST_LOGIN = env.bool("UPDATE_LAST_LOGIN")
UPDATE_LAST_LOGIN_INTERVAL_MINUTES = env.int("UPDATE_LAST_LOGIN_INTERVAL_MINUTES")
```

UPDATE_LAST_LOGIN["UPDATE_INTERVAL"] drops the MINUTES from the variable name and thus makes it less clear what is the unit. One could argue in favor of keeping the MINUTES in the variable name here with e.g. Zen of Python's "Explicit is better than implicit".

palvelutarjotin/tests/test_oidc.py
```diff
@@ -1,13 +1,31 @@
...
+def mocked_jwt_request(encoded_jwt):
+    with mock.patch(
+        "helusers.oidc.ApiTokenAuthentication.decode_jwt",
+        return_value=JWTClaims(
+            jwt.decode(encoded_jwt, options={"verify_signature": False}), None
+        ),
+    ):
+        yield RequestFactory().get(
+            "/fake-path",
+            HTTP_AUTHORIZATION="Bearer {}".format(encoded_jwt),
```
karisal-anders, line 25: Using format string would make this more concise `f"Bearer {encoded_jwt}"`, but it's fine to leave it this way too.

palvelutarjotin/oidc.py
```diff
@@ -19,7 +22,30 @@ def validate_claims(self, id_token):
...
+        minutes_since_last_update = None
+        if user.last_login:
+            minutes_since_last_update = (
+                timezone.now() - user.last_login
+            ).total_seconds() / 60
+        if (
+            minutes_since_last_update is None
+            or minutes_since_last_update > settings.UPDATE_LAST_LOGIN["UPDATE_INTERVAL"]
+        ):
+            update_last_login(None, user)
```
karisal-anders, line 43: One has to know or look up the function signature to know what the first parameter `None` means here. Using keyword arguments here would add clarity IMHO and could be argued in favor of with Zen of Python's "Explicit is better than implicit":
```python
update_last_login(sender=None, user=user)
```
What do you think?

karisal-anders (commented):
Left some comments. The tests look good now IMHO, maybe adding some rationale to e.g. FixLastLoginMixin's documentation about why the the mixin is needed would be good.

nikomakela:
Added a commit to fix some newly created tests that did not mock the get event from linked events connection, so they actually used a real one!

palvelutarjotin/oidc.py
nikomakela: Changed

palvelutarjotin/tests/test_oidc.py
nikomakela: Changed

palvelutarjotin/settings.py
nikomakela: Did a change

palvelutarjotin/oidc.py
```diff
@@ -19,7 +22,31 @@ def validate_claims(self, id_token):
...
+        """
+        minutes_since_last_update = None
+        if user.last_login:
+            minutes_since_last_update = (
+                timezone.now() - user.last_login
+            ).total_seconds() / 60
+        if (
+            minutes_since_last_update is None
+            or minutes_since_last_update
+            > settings.UPDATE_LAST_LOGIN["UPDATE_INTERVAL_MINUTES"]
```
karisal-anders, line 42: This way black formats Python code is horrible—meaning the indentation levels being all identical also for the or clause's innards—to me but it's probably still PEP8 compliant. I mean you can leave it like this because that's how black does it and we're using black for formatting but I would add parentheses around the or i.e. `or (minutes_since_last_update > settings.UPDATE_LAST_LOGIN["UPDATE_INTERVAL_MINUTES"])` to make black format the code more readably.

palvelutarjotin/oidc.py
```diff
@@ -19,7 +22,30 @@ def validate_claims(self, id_token):
         super().validate_claims(id_token)
 
 
-class GraphQLApiTokenAuthentication(FixTunnistamoMixin, ApiTokenAuthentication):
+class FixLastLoginMixin:
+    def update_last_login_after_interval(self, user):
+        """
+        Update the user object last_login field if enough time has passed.
+        The setting UPDATE_LAST_LOGIN_INTERVAL determines whether or not
```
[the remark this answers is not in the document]
karisal-anders: The documentation of the function is still out of date, please would you update it? Adding some rationale—like in `FixTunnistamoMixin` there's a brief rationale of why the functionality exists—would be a good addition IMHO but you can decide what to do with that.

karisal-anders (commented):
One snapshot test case is failing and one docstring is out of date but otherwise LGTM.

palvelutarjotin/oidc.py
nikomakela: black is black. Let's say that my interest to this PR is currently in zero, so I just take what the formatter gives.

PR #311, opened 2023-04-25
Security fixes to enrolment data fetching

nikomakela:
Based on https://github.com/City-of-Helsinki/palvelutarjotin/pull/307.

This PR is to check and improve the security related to following:

1. The enrolment data should not be available publicly
2. The enrolment data from other organisation should not be available for a staff member.


nikomakela:
A project like this could be added to the dependencies https://pypi.org/project/django-graphene-permissions/ to bring a better maintained alternative solution to all this.

terovirtanen:
**kultus-api branch is deployed to platta: https://kultus-api-pr311.dev.hel.ninja** :rocket::rocket::rocket:

occurrences/models.py
```diff
@@ -495,6 +497,23 @@ def with_enrolments_count(self, use_name_only=False):
...
 
+    def user_can_view(self, user: User):
+        if not user.person:
+            return self.none
+        organisation_ids = [
+            entry[0] for entry in user.person.organisations.values_list("id")
+        ]
+        return self.with_organisation_ids().filter(
+            organisation_ids__contains=organisation_ids
+        )
```
karisal-anders, line 508: I tried to see what query this generates but was unable as running ```python
str(self.with_organisation_ids().filter(organisation_ids__contains=organisation_ids).query)
```
here gives
```python
*** django.core.exceptions.FieldError: Cannot resolve expression type, unknown output_field
```

I was trying to check whether this filter works so that the list on the left has to contain all of the values from the list on the right to pass through or not. I would think so, so then a question arises: Is this necessary or is this a too strict check? Should we let it through if at least a single value from the list on the right is found in the list on the left? How do you see this?

karisal-anders: Perhaps PostgreSQL specific [overlap](https://docs.djangoproject.com/en/3.2/ref/contrib/postgres/fields/#overlap) functionality could be used here:

```python
        return self.with_organisation_ids().filter(
            organisation_ids__overlap=organisation_ids
        )
```
?

Then this would pass AFAIK if at least a single value from the list on the right is found in the list on the left.

karisal-anders: This boils down to is the current filtering the correct one or does making it less strict actually let through data that we shouldn't let the user see?

nikomakela: I mean, I'm OK to change it, I just don't understand the difference here. Is the problem that the aggregated value is not  an arrayField?

nikomakela:
I tested the changes against both the UIs, **while I had still not updated their schemas**. They seem to work OK.

First I setup a local environment with a local API, a local Tunnistamo and a local LinkedEvents.

1. I could create a new event event from the admin-UI 
image
2. I could see the new event in the admin-UI list of future events 
image
3. I could update the already published event 
image
4. I could see the newly created event in the public UI 
image 
image
5. I could enrol to the event occurrence 
image
image
6. I could see a list of places populated in a unit field 
image
7. I could see the enrolment in the admin-UI 
image
8. I could approve the enrolment
image

**No warnings were raised when I went through all these steps!**

occurrences/models.py
nikomakela: Yea, so, the overlap really is the better option and it's now changed. Overlap means that "any of the items occurs" and contains meant that "all needs to occur".

nikomakela: Nice catch @karisal-anders !

occurrences/tests/test_models.py
```diff
@@ -172,6 +172,65 @@ def test_study_group_with_enrolment_count_by_name(
...
+    person.organisations.add(organisation)
+    organisation_enrolments = EnrolmentFactory.create_batch(
+        5, occurrence__p_event__organisation=organisation
+    )
+    organisation_groups = [
+        enrolment.study_group for enrolment in organisation_enrolments
+    ]
+    StudyGroupFactory.create_batch(5)
+    StudyGroup.objects.all().count() == 10
+    StudyGroup.objects.user_can_view(person.user).count() == len(organisation_groups)
```
karisal-anders, line 231: asserts? (See Sonarcloud [report](https://sonarcloud.io/project/issues?resolved=false&types=BUG&sinceLeakPeriod=true&pullRequest=311&id=City-of-Helsinki_kultus-api))

karisal-anders (approved):
I think this is an improvement to the situation, things that could be improved:
- the **asserts** in `test_study_group_user_can_view` **should be adde**d before merging IMO (See [comment](https://github.com/City-of-Helsinki/palvelutarjotin/pull/311/files#r1177851298))
- maybe add tests for cases where organisation does exist but is different so the filtering should still return an empty set, now there are tests where organisation does not exist at all and where the same organisation exists but no test where organisation exists and is different

terovirtanen:
**kultus-api branch is deployed to platta: https://kultus-api-pr311.dev.hel.ninja** :rocket::rocket::rocket:

PR #341, opened 2024-02-21
 add keyword_OR_set event query parameters

nikomakela:
PT-1591.
Added keyword_OR_set1, keyword_OR_set2 and keyword_OR_set3 for the keywords of target groups, categories and activities / additional criterias.

### DIFF
--- a/graphene_linked_events/tests/test_api.py
+++ b/graphene_linked_events/tests/test_api.py
@@ -647,6 +647,9 @@ def test_pevent_preadded_with_test_events_p_event_relations():
         ("keywordNot", "keyword!"),
         ("allOngoingAnd", "all_ongoing_AND"),
         ("allOngoingOr", "all_ongoing_OR"),
+        ("keywordOrSet1", "keyword_OR_set1"),
+        ("keywordOrSet2", "keyword_OR_set2"),
+        ("keywordOrSet3", "keyword_OR_set3"),
     ],
 )
 def test_resolve_events_unsupported_parameter_mapping(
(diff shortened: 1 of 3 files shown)


terovirtanen:
**KULTUS-API branch is deployed to platta: https://kultus-pr341.api.dev.hel.ninja** :rocket::rocket::rocket:

graphene_linked_events/tests/test_api.py
```diff
@@ -647,6 +647,9 @@ def test_pevent_preadded_with_test_events_p_event_relations():
         ("keywordNot", "keyword!"),
         ("allOngoingAnd", "all_ongoing_AND"),
         ("allOngoingOr", "all_ongoing_OR"),
+        ("keywordOrSet1", "keyword_OR_set1"),
+        ("keywordOrSet2", "keyword_OR_set2"),
+        ("keywordOrSet3", "keyword_OR_set3"),
     ],
 )
 def test_resolve_events_unsupported_parameter_mapping(
```
karisal-anders, line 655: This was not changed in this PR but just curious why is this function named like it is? Unsupported parameter mapping? Why unsupported? Isn't it clearly supported as they're being tested that the mapping happens in a certain way? Maybe the naming of this function could be made better to signify what's going on here, it is unclear IMHO.

karisal-anders (approved):
LGTM, one of the names of the test functions a bit unclear but that wasn't introduced in this PR

graphene_linked_events/tests/test_api.py
nikomakela: I think it tests something like, if you would add custom parameters to the URL, it would not break the LinkedEvents query.

PR #343, opened 2024-03-08
Fix slowness in the certain admin pages

jakezu:
The fix involved correcting the definition of certain form fields as readonly fields while designating others as autocomplete fields.

### DIFF
--- a/occurrences/admin.py
+++ b/occurrences/admin.py
@@ -34,10 +34,32 @@ class OccurrenceInline(admin.TabularInline):
     model = Occurrence.study_groups.through
     extra = 0
 
+    def has_change_permission(self, request, obj=None):
+        return False
+
+    # Without setting these add and delete permissions to false, the page loads
+    # slowly due to a large number of SQL queries
+    def has_add_permission(self, request, obj=None):
+        return False
+
+    def has_delete_permission(self, request, obj=None):
+        return False
+
 
 class EventQueueEnrolmentInline(admin.TabularInline):
     model = EventQueueEnrolment
     extra = 0
+
+    def has_change_permission(self, request, obj=None):
+        return False
+
+    # Without setting these add and delete permissions to false, the page loads
+    # slowly due to a large number of SQL queries
+    def has_add_permission(self, request, obj=None):
+        return False
+
+    def has_delete_permission(self, request, obj=None):
+        return False
 
 
 class EnrolmentInline(admin.TabularInline):
@@ -45,11 +67,17 @@ class EnrolmentInline(admin.TabularInline):
     extra = 0
     exclude = ("person_deleted_at",)
 
+    def has_change_permission(self, request, obj=None):
+        return False
+
 
 class StudyGroupInline(admin.TabularInline):
     model = StudyGroup
     extra = 0
     exclude = ("person_deleted_at",)
+
+    def has_change_permission(self, request, obj=None):
+        return False
 
 
 class HasUnitIdStudyGroupListFilter(admin.SimpleListFilter):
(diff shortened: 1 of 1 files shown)


terovirtanen:
**KULTUS-API branch is deployed to platta: https://kultus-pr343.api.dev.hel.ninja** :rocket::rocket::rocket:

karisal-anders (approved):

occurrences/admin.py
```diff
@@ -34,23 +34,51 @@ class OccurrenceInline(admin.TabularInline):
...
+    def has_change_permission(self, request, obj=None):
+        return False
+
+    # Without setting these add and delete permissions to false, the page loads
+    # slowly due to a large number of SQL queries
+    def has_add_permission(self, request, obj=None):
+        return False
+
+    def has_delete_permission(self, request, obj=None):
+        return False
```
nikomakela, line 62: The enrolments are something that we really might have to edit from the Django admin.

```suggestion
    readonly_fields = ['person_deleted_at']
    autocomplete_fields = ['person', 'p_event']
```

occurrences/admin.py
```diff
@@ -34,23 +34,51 @@ class OccurrenceInline(admin.TabularInline):
...
+
 
 class EnrolmentInline(admin.TabularInline):
     model = Enrolment
     extra = 0
     exclude = ("person_deleted_at",)
 
+    def has_change_permission(self, request, obj=None):
+        return False
+
```
nikomakela, line 72: I think that here in `EnrolmentInline` this `has_change_permission` is actually fine, but I would still use the `autocomplete_fields` also when adding the enrolments, because without the search feature, the dropdown is just useless.

```suggestion
    autocomplete_fields = ['study_group', 'occurrence']
    
    def has_change_permission(self, request, obj=None):
        return False

```

occurrences/admin.py
```diff
@@ -34,23 +34,51 @@ class OccurrenceInline(admin.TabularInline):
...
 
 class StudyGroupInline(admin.TabularInline):
     model = StudyGroup
     extra = 0
     exclude = ("person_deleted_at",)
 
+    def has_change_permission(self, request, obj=None):
+        return False
+
```
nikomakela, line 81: I think that here in `StudygroupInline` this `has_change_permission` is fine, but sure there can be cases that we may need to change the name of the group if it is misspelled or something. Since I don't think that fetching the permissions should be a big deal or a deal at all, I would'nt restrict editing.
```suggestion
```

occurrences/admin.py
```diff
@@ -123,6 +163,21 @@ class OccurrenceAdmin(admin.ModelAdmin):
     exclude = ("id",)
     list_filter = ["start_time", "end_time", "seat_type", "cancelled"]
     search_fields = ["p_event__linked_event_id"]
+    readonly_fields = ["p_event"]
+    autocomplete_fields = ["contact_persons"]
```
nikomakela, line 167: **The `p_event` cannot be readonly in `OccurrenceAdmin` or you cannot add a new occurrence!**. Move it to `autocomplete_fields` instead.

```suggestion
    autocomplete_fields = ["p_event", "contact_persons"]
```

nikomakela (changes requested):
**The `p_event` field cannot be a readonly field in the `OccurrenceAdmin`!**

I wouldn't use the permissions-methods overriding that much either, because I don't think the issue is in fetching the permissions, but in rendering the edit form fields.

occurrences/admin.py
```diff
@@ -34,23 +34,51 @@ class OccurrenceInline(admin.TabularInline):
...
+        return False
+
+    # Without setting these add and delete permissions to false, the page loads
+    # slowly due to a large number of SQL queries
+    def has_add_permission(self, request, obj=None):
+        return False
+
+    def has_delete_permission(self, request, obj=None):
+        return False
+
```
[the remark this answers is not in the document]
jakezu: Fixed as suggested.

occurrences/admin.py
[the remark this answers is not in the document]
jakezu: That's right. I misinterpreted that the problem specifically lay in the large number of SQL queries.

occurrences/admin.py
jakezu: I added `autocomplete_fields` but removed `has_change_permission` because setting the `autocomplete_fields` fixed the original issue. It's not necessary to restrict editing capability here either.

occurrences/admin.py
jakezu: Sure. It's not necessary to restrict editing capability here either.

occurrences/admin.py
jakezu: Fixed as suggested.

jakezu (commented):
I made changes that do not break editing capabilities. There was also no need to leave some readonly fields.

nikomakela (approved):

nikomakela:
Note the coding style error that must be addressed

terovirtanen:
**KULTUS-API branch is deployed to platta: https://kultus-pr343.api.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
**KULTUS-API branch is deployed to platta: https://kultus-pr343.api.dev.hel.ninja** :rocket::rocket::rocket:

PR #365, opened 2024-09-25
Resolve if pEvent has any spaces left for enrolment

nikomakela:
PT-1668.

Also: 
- feat: occurrences are listed in inline component in palvelutarjotin event admin
- feat: improve the performance of admin site with auto complete fields and inline components
- refactor: move upcoming events annotation to query set. Add a new `with_next_occurrence` annotation to PalvelutarjotinEvent queryset
to do the job that was done in the upcoming events query endpoint.

### DIFF
--- a/graphene_linked_events/tests/test_api.py
+++ b/graphene_linked_events/tests/test_api.py
@@ -31,8 +31,12 @@ from graphene_linked_events.tests.mock_data import (
 from graphene_linked_events.tests.utils import MockResponse
 from graphene_linked_events.utils import retrieve_linked_events_data
 from occurrences.event_api_services import update_event_to_linkedevents_api
-from occurrences.factories import OccurrenceFactory, PalvelutarjotinEventFactory
-from occurrences.models import PalvelutarjotinEvent
+from occurrences.factories import (
+    EnrolmentFactory,
+    OccurrenceFactory,
+    PalvelutarjotinEventFactory,
+)
+from occurrences.models import Occurrence, PalvelutarjotinEvent
 from palvelutarjotin.consts import API_USAGE_ERROR, DATA_VALIDATION_ERROR
 from palvelutarjotin.exceptions import ApiBadRequestError, ObjectDoesNotExistError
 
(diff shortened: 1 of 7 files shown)


terovirtanen:
**KULTUS-API branch is deployed to platta: https://kultus-pr365.api.dev.hel.ninja** :rocket::rocket::rocket:

occurrences/models.py
```diff
@@ -105,6 +105,17 @@ def delete_contact_info(self, now=None):
             contact_info_deleted_at=now,
         )
 
+    def with_next_occurrence(self):
+        next_occurrences = Occurrence.objects.filter(
+            p_event=OuterRef("pk"), start_time__gte=timezone.now(), cancelled=False
+        ).order_by("start_time")
+
+        return self.annotate(
+            next_occurrence_start_time=Subquery(
```
karisal-anders, line 114: How about `with_next_occurrence_start_time` instead of `with_next_occurrence`? That would specify exactly what is annotated, and still it's not yet overly long IMHO.

graphene_linked_events/tests/test_api.py
```diff
@@ -2091,3 +2095,147 @@ def test_linkedevents_api_update_errors(status_code, error_cls):
...
+        p_event=p_event,
+        start_time=timezone.now() - timedelta(days=1),
+        cancelled=False,
+        amount_of_seats=2,
+    )
+    # test with seat_Type set to OCCURRENCE_SEAT_TYPE_ENROLMENT_COUNT
+    occurrence1 = OccurrenceFactory(
+        p_event=p_event,
+        start_time=timezone.now() + timedelta(days=1),
+        seat_type=Occurrence.OCCURRENCE_SEAT_TYPE_CHILDREN_COUNT,
```
karisal-anders, line 2138: Comment and code for `seat_type` are out-of-sync

occurrences/models.py
```diff
@@ -964,7 +1004,7 @@ def approve(self, send_notification=True, custom_message: Optional[str] = None):
         Raises:
             EnrolmentNotEnoughCapacityError: Not enough space for the group
         """
-        if self.occurrence.seats_taken > self.occurrence.amount_of_seats:
+        if not self.occurrence.has_space_left():
```
karisal-anders, line 1007: `a > b` here has been changed to `not(a < b)` i.e. `a >= b` so the implementation has changed. Is this intentional?

nikomakela: Good catch, because I was first going to say that it was intentional, but then again, I started to think where it was used and how that function really should check it and I came to conclusion, that this line of code really needs to test whether the change that was already made still fits in the defined limits. So yes, this would have changed the feature so that 1 empty place would have always be needed. I'll write a new test for this too.

nikomakela: I reverted this line's code change and improved an old test related to this so that it tests this case too.

terovirtanen:
**KULTUS-API branch is deployed to platta: https://kultus-pr365.api.dev.hel.ninja** :rocket::rocket::rocket:

PR #370, opened 2024-10-28
Deactivated person should not be listed from API

nikomakela:
PT-1804.

There are 2 kind of persons in the API: Those that acts as event
providers and are linked to some admin users, but also those that leads
a study group or makes an enrolment to the event. Both types are casted
as PersonNode, but only one of them has a user linked to it. However,
there is no case where we should currently include the deactivated user
instances, so those can always be excluded.

NOTE: After the authorization service is changed from the Tunnistamo to
the Helsinki Profile Keycloak, there will be duplicate user instances
for the admin users. Then, the clearest way to separate the old
Tunnistamo accounts from the active ones, is to mark them deactivated
and at that point, we don't want to have them listed in the event
contact person list or anywhere else.

### DIFF
--- a/organisations/schema.py
+++ b/organisations/schema.py
@@ -126,11 +126,13 @@ class CreateMyProfileMutation(graphene.relay.ClientIDMutation):
         person.notify_myprofile_creation()
         return CreateMyProfileMutation(my_profile=person)
 
+    @staticmethod
     def _set_organisation_proposals(organisation_proposals_data, person):
         if organisation_proposals_data and person:
             for organisation_proposal_data in organisation_proposals_data:
                 person.organisationproposal_set.create(**organisation_proposal_data)
 
+    @staticmethod
     def _set_person_organisations(organisation_ids, person):
         if organisation_ids and person:
             for org_global_id in organisation_ids:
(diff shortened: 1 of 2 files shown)


nikomakela:
For some reason, the automated checks are not running as expected. They don't start at all. The same issue was in Kukkuu API: https://github.com/City-of-Helsinki/kukkuu/pull/430 and https://dev.azure.com/City-of-Helsinki/kukkuu/_build/results?buildId=206206&view=logs&j=d319fce4-cb46-5c9b-3d68-d529030f0183&t=2120c938-01a3-50af-fdbe-41d2d4cfb9f6&l=533. The Kukkuu is having some issues with Postgre!

- [ ] Fix the issue with pipelines and automated checks

karisal-anders (approved):

karisal-anders:
The branch name is so long that the PostgreSQL database's name in the review environment goes over maximum of 64 characters and fails:
> AttributeError: 'NoneType' object has no attribute 'info'
ImproperlyConfigured: The database name 'R20241029_kultusapi66ed2dev_pt_1804_api_should_not_list_deactivated_persons' (75 characters) is longer than PostgreSQL's limit of 63 characters. Supply a shorter NAME in settings.DATABASES.
(14 additional frame(s) were not displayed)

nikomakela:
Closing this for https://github.com/City-of-Helsinki/palvelutarjotin/pull/371 with a shorter branch name

PR #371, opened 2024-10-29
Deactivated person should not be listed from API

nikomakela:
A duplicate for https://github.com/City-of-Helsinki/palvelutarjotin/pull/370 with a shorter branch name, because it was an issue for infra.

----

PT-1804.

There are 2 kind of persons in the API: Those that acts as event
providers and are linked to some admin users, but also those that leads
a study group or makes an enrolment to the event. Both types are casted
as PersonNode, but only one of them has a user linked to it. However,
there is no case where we should currently include the deactivated user
instances, so those can always be excluded.

NOTE: After the authorization service is changed from the Tunnistamo to
the Helsinki Profile Keycloak, there will be duplicate user instances
for the admin users. Then, the clearest way to separate the old
Tunnistamo accounts from the active ones, is to mark them deactivated
and at that point, we don't want to have them listed in the event
contact person list or anywhere else.

### DIFF
--- a/organisations/schema.py
+++ b/organisations/schema.py
@@ -126,11 +126,13 @@ class CreateMyProfileMutation(graphene.relay.ClientIDMutation):
         person.notify_myprofile_creation()
         return CreateMyProfileMutation(my_profile=person)
 
+    @staticmethod
     def _set_organisation_proposals(organisation_proposals_data, person):
         if organisation_proposals_data and person:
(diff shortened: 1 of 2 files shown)


karisal-anders (approved):

terovirtanen:
**KULTUS-API branch is deployed to platta: https://kultus-pr371.api.dev.hel.ninja** :rocket::rocket::rocket:

PR #384, opened 2025-03-26
Audit logging

nikomakela:
PT-1851.

Add django-auditlog and django-auditlog-extra to requirements.

The User deletion (especially GDPR deletion) needed some special handling related to audit logging, because the actor cannot be set to same object that is being deleted, because of foreign key constaint violation.

Also, fixed: "Running unit tests with `pytest` failed in cases when `Occurence` or
`EventQueueEnrolment` did not have `p_event` set.".


karisal-anders:
@nikomakela `gdpr/tests/test_gdpr_api.py::test_delete_profile_data_from_gdpr_api[simplest]` test case [is failing](https://dev.azure.com/City-of-Helsinki/kultus/_build/results?buildId=234586&view=logs&jobId=8267e077-3a31-5bba-93c5-4b3e55896ad0&j=8267e077-3a31-5bba-93c5-4b3e55896ad0&t=45ce9a3b-1c36-5c65-31cd-c5d7e373a2ce) in the pipeline:
```
=================================== FAILURES ===================================
_______________ test_delete_profile_data_from_gdpr_api[simplest] _______________

test_scope = 'simplest'
snapshot = <snapshottest.pytest.PyTestSnapshotTest object at 0x7fd7d53a6f60>
gdpr_api_client = <rest_framework.test.APIClient object at 0x7fd7d53a56d0>
requests_mock = <requests_mock.mocker.Mocker object at 0x7fd7d55e06e0>
mock_get_event_data = None

    @pytest.mark.django_db(reset_sequences=True, transaction=True)
    @pytest.mark.parametrize(
        "test_scope",
        ["simplest", "most_complex"],
    )
    def test_delete_profile_data_from_gdpr_api(
        test_scope, snapshot, gdpr_api_client, requests_mock, mock_get_event_data
    ):
        # Clear the LogEntry table before the test
        LogEntry.objects.all().delete()
        assert LogEntry.objects.count() == 0
        # Create a new user
        user = UserFactory()
    
        if test_scope == "most_complex":
            _test_person_with_all_data_relations_populated(user, assert_counts=True)
    
        (
            enrolment_count,
            study_group_count,
            event_queue_enrolment_count,
        ) = original_counts = _get_db_table_counts_of_person_related_tables()
    
        auth_header = get_api_token_for_user_with_scopes(
            user, [settings.GDPR_API_DELETE_SCOPE], requests_mock
        )
        gdpr_api_client.credentials(HTTP_AUTHORIZATION=auth_header)
        response = gdpr_api_client.delete(
            reverse(
                "helsinki_gdpr:gdpr_v1",
                kwargs={settings.GDPR_API_MODEL_LOOKUP: user.uuid},
            )
        )
        assert response.status_code == 204
        with pytest.raises(User.DoesNotExist):
            User.objects.get(username=user.username)
    
        if test_scope == "most_complex":
            all_count = LogEntry.objects.count()
            assert all_count > 2
            created = LogEntry.objects.filter(action=LogEntry.Action.CREATE)
            deleted = LogEntry.objects.filter(action=LogEntry.Action.DELETE)
            assert created.count() == all_count - deleted.count()
            assert deleted.count() == 1
            assert deleted.first().object_id == user.id
        else:
            assert LogEntry.objects.count() == 2
            (created_log, deleted_log) = list(LogEntry.objects.all())
>           assert created_log.action == LogEntry.Action.CREATE
E           AssertionError: assert 2 == 0
E            +  where 2 = <LogEntry: Deleted Black Alexis (joshuajohnson@example.com)>.action
E            +  and   0 = <class 'auditlog.models.LogEntry.Action'>.CREATE
E            +    where <class 'auditlog.models.LogEntry.Action'> = LogEntry.Action

gdpr/tests/test_gdpr_api.py:271: AssertionError
```

nikomakela:
> assert LogEntry.objects.count() == 2

When I run only that one test case, it passes. If I ran only that one test suite, it passes. If I ran all the tests together, it does not pass. Does not make any sense. Also, it is checking the data counts many times and it seems that the right data really is there, but it jsut asserts it wrong.

I guess the count query might also need a "filter" too, so I also tried with

```
assert LogEntry.objects.all().count() == 2
```

However, it didn't make any difference. I need to investigate further.

nikomakela:
Converted to draft, because I realized that a GraphQL endpoint access logging implementation is still missing.

- [ ] Add access logging to GraphQL endpoints

terovirtanen:
**KULTUS-API branch is deployed to platta: https://kultus-pr384.api.dev.hel.ninja** :rocket::rocket::rocket:

README.md
```diff
@@ -276,6 +277,24 @@ allowed_chars = string.ascii_lowercase + string.digits + "!@#$%^&*(-_=+)"
 To view the API documentation, in DEBUG mode visit http://localhost:8081/graphql and checkout the
 `Documentation Explorer` section
 
+## Audit logging
+
+Audit logging is implemented with `django-auditlog`, but it has some extended features applied with [auditlog_extra](https://github.com/City-of-Helsinki/django-auditlog-extra) -app.
+
+The configuration to define which models are in the scope of the audit logging can be found in [auditlog_settings.py](./palvelutarjotin/auditlog_settings.py).
```
karisal-anders, line 284:
```suggestion
The configuration that defines which models are in the scope of the audit logging can be found in [auditlog_settings.py](./palvelutarjotin/auditlog_settings.py).
```

occurrences/admin.py
```diff
@@ -246,7 +249,7 @@ class VenueCustomDataAdmin(TranslatableAdmin):
 
 
 @admin.register(PalvelutarjotinEvent)
-class PalvelutarjotinEventAdmin(admin.ModelAdmin):
+class PalvelutarjotinEventAdmin(AuditlogAdminViewAccessLogMixin, admin.ModelAdmin):
```
karisal-anders, line 252: As PalvelutarjotinEventAdmin's list_display contains get_contact_email which shows GDPR data (i.e. an email) wouldn't it make sense to log the list views because of this?
```suggestion
class PalvelutarjotinEventAdmin(AuditlogAdminViewAccessLogMixin, admin.ModelAdmin):
    enable_list_view_audit_logging = True
```

occurrences/admin.py
```diff
@@ -312,7 +315,7 @@ def get_contact_email(self, obj):
 
 
 @admin.register(EnrolleePersonalData)
-class EnrolleePersonalDataAdmin(admin.ModelAdmin):
+class EnrolleePersonalDataAdmin(AuditlogAdminViewAccessLogMixin, admin.ModelAdmin):
```
karisal-anders, line 318: As EnrolleePersonalDataAdmin's list_display contains name, phone_number and email_address which are GDPR data, wouldn't it make sense to log the views because of this?
```suggestion
class EnrolleePersonalDataAdmin(AuditlogAdminViewAccessLogMixin, admin.ModelAdmin):
    enable_list_view_audit_logging = True
```

occurrences/admin.py
```diff
@@ -193,7 +196,7 @@ def linked_event_id(self, obj):
 
 
 @admin.register(EventQueueEnrolment)
-class EventQueueEnrolmentAdmin(admin.ModelAdmin):
+class EventQueueEnrolmentAdmin(AuditlogAdminViewAccessLogMixin, admin.ModelAdmin):
```
karisal-anders, line 199: As EventQueueEnrolmentAdmin's list_display contains get_person which shows the person's name (See screenshot for example) which is GDPR data, wouldn't it make sense to log the list views because of this?
```suggestion
class EventQueueEnrolmentAdmin(AuditlogAdminViewAccessLogMixin, admin.ModelAdmin):
    enable_list_view_audit_logging = True
```
image

occurrences/models.py
```diff
@@ -435,7 +435,10 @@ def delete(self, *args, **kwargs):
         self.__post_delete_unpublish_event()
 
     def __str__(self):
-        return f"{self.p_event.linked_event_id} {self.start_time} {self.place_id}"
+        if self.p_event:
+            return f"Event {self.p_event.linked_event_id} at {self.place_id} ({self.start_time.strftime('%Y-%m-%d %H:%M')})"  # noqa: E501
+        else:
+            return f"Event at {self.place_id} ({self.start_time.strftime('%Y-%m-%d %H:%M')})"  # noqa: E501
```
karisal-anders, line 441: One alternative to get rid of the long lines:
```suggestion
        event = f"Event {self.p_event.linked_event_id}" if self.p_event else "Event"
        return (
            f"{event} at {self.place_id} ({self.start_time.strftime('%Y-%m-%d %H:%M')})"
        )
```

organisations/admin.py
```diff
@@ -74,7 +75,7 @@ def __init__(self, *args, **kwargs):
 
 
 @admin.register(OrganisationProposal)
-class OrganisationProposalAdmin(admin.ModelAdmin):
+class OrganisationProposalAdmin(AuditlogAdminViewAccessLogMixin, admin.ModelAdmin):
```
karisal-anders, line 78: As OrganisationProposalAdmin's list_display contains name which is the applicant's name and GDPR data, wouldn't it make sense to log the list views because of this?
```suggestion
class OrganisationProposalAdmin(AuditlogAdminViewAccessLogMixin, admin.ModelAdmin):
    enable_list_view_audit_logging = True
```

organisations/models.py
```diff
@@ -112,8 +114,18 @@ class Meta:
...
+        # Make a log entry for the deletion
+        LogEntry.objects.log_create(
+            self, force_log=True, action=LogEntry.Action.DELETE
+        ).save()
+        # Delete the user without creating a auditlog entry,
+        # because the log entry is already created above and
+        # there would be a foreign key constraint violation
+        # if the log entry is not created before the user is deleted.
+        with disable_auditlog():
+            return super().delete(*args, **kwargs)
```
karisal-anders, line 128: Can there be any other related objects that are cascade deleted here? If there can be, then shouldn't their deletions still be audit logged here?

karisal-anders (commented):
Please take a look at the comments I left—if you'd like we can have a live talk if there's anything you'd want to discuss related to these.

occurrences/admin.py
nikomakela: it would mean that every time an event card is shown in the UI, something would be written in the logs. Even then when using public UI and the user is not even authenticated.

nikomakela: Another option would be to remove the email column (and other gdpr columns) from the list view.

occurrences/admin.py
nikomakela: I understand your logic, but I still would not enable the `enable_list_view_audit_logging` since the person name is the only information there. It's shared information anyway and it would fool the auditlog readers, that somebody really saw all the contact information, etc. Writing access log of these would just spam the auditlog, wouldn't it?

organisations/admin.py
nikomakela: This is just spam to auditlog. Organisation names are public anyway. It's only admins who are reading them from django admin site.

image

organisations/models.py
nikomakela: That was my concern too, but the GDPR related tests should check these. The related data is "anonymized"  before calling the user delete. To be honest, I could not find any other way to do this.

occurrences/models.py
nikomakela: If using the noqa: E501 is a problem here, I would be more willing to replace this with string format-command instead.

occurrences/admin.py
nikomakela: Yea, this could be added, since there are so many GDPR columns.

occurrences/admin.py
nikomakela: This contact information is also public information that is shown also in the public UI's event cards and details. Users are anonymous (not authenticated) when they are using the public UI. I would not like to spam the auditlog only because of this public information.

terovirtanen:
**KULTUS-API branch is deployed to platta: https://kultus-pr384.api.dev.hel.ninja** :rocket::rocket::rocket:

occurrences/models.py
karisal-anders: Just thought I'd get rid of the `# noqa: E501`'s and long lines. If you have other ideas how to make do without them, go ahead.

occurrences/admin.py
nikomakela: How about masking the person column?

occurrences/admin.py
nikomakela: How about masking the email column?

terovirtanen:
**KULTUS-API branch is deployed to platta: https://kultus-pr384.api.dev.hel.ninja** :rocket::rocket::rocket:

occurrences/admin.py
karisal-anders: > This contact information is also public information that is shown also in the public UI's event cards and details.

That **functionality** seems to be **bugged** at the moment i.e. they seem to be never shown. I looked into it a bit and the even when a contact person is set it comes through as null in the event graphql query, which makes the <OrganisationInfo> component not show the contact details.

But, the contact person's phone number and email at least can be seen publicly through the data fetch. Because the contact person's phone number and email are currently publicly available and the contact person's name would also be if the functionality worked I'd say this is ok not to add the list view audit logging here.

terovirtanen:
**KULTUS-API branch is deployed to platta: https://kultus-pr384.api.dev.hel.ninja** :rocket::rocket::rocket:

occurrences/admin.py
karisal-anders: Created ticket [PT-1898](https://helsinkisolutionoffice.atlassian.net/browse/PT-1898) related to this.

organisations/models.py
karisal-anders: Do you mean `test_delete_profile_data_from_gdpr_api`? So is it so that the anonymization changes to related enrolments, study groups and event queue enrolments are not logged here now?

terovirtanen:
**KULTUS-API branch is deployed to platta: https://kultus-pr384.api.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
**KULTUS-API branch is deployed to platta: https://kultus-pr384.api.dev.hel.ninja** :rocket::rocket::rocket:

common/mixins.py
```diff
@@ -1,3 +1,86 @@
+from auditlog.context import set_actor
+from auditlog.signals import accessed
+from rest_framework import mixins
+
+
+class LogListAccessMixin(mixins.ListModelMixin):
```
karisal-anders, line 6: Looks generic enough to be perhaps be worthy to include in django-auditlog-extra itself, or what do you think?

occurrences/admin.py
karisal-anders: After discussion this point can be left as is, or masked, if wanted to, and can be possibly iterated on later (not in this PR).

organisations/admin.py
karisal-anders: After discussion this point can be left as is, or masked, if wanted to, and can be possibly iterated on later (not in this PR).

organisations/models.py
karisal-anders: This is something that would be good to look into / check what's going on here really. And the GDPR tests should be fixed as they're at least partially broken at the moment (i.e. the snapshot contains GDPR data when it should be anonymized). But both of these things can be done outside of this PR (or in this PR if you want to, but not necessary).

reports/views.py
```diff
@@ -38,6 +41,17 @@ def naive_datetime_to_tz_aware(datetime: str) -> str:
     return datetime + "T00:00:00.000Z" if datetime else None
 
 
+class LogAccessMixin:
```
karisal-anders, line 44: What do you think, would this be worthy to add to django-auditlog-extra for additional DRF support?

terovirtanen:
**KULTUS-API branch is deployed to platta: https://kultus-pr384.api.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
**KULTUS-API branch is deployed to platta: https://kultus-pr384.api.dev.hel.ninja** :rocket::rocket::rocket:

nikomakela:
> Looks ok to merge with the given comments (i.e. see the comments about the disable_auditlog use & at least partially broken GDPR tests)

@karisal-anders I now added some improvements that you could check again.

terovirtanen:
**KULTUS-API branch is deployed to platta: https://kultus-pr384.api.dev.hel.ninja** :rocket::rocket::rocket:

reports/views.py
nikomakela: A view mixin is not a good solution at all, because it can only take care of the first level of the object. What if a detail view or a list view shows some embedded objects or object relations? A generic view mixin cannot know about those. We can provide a get_queryset method or get_context_data method through a mixin, but the context scope of it is still unknown.

A generic reusable mixin would make more sense if it would be on serialization level. Some kind of object serializer is called everytime, no matter how deep we go.  It's easier to provide a generic mixin in GraphQL scope, because there a get_node is called for every level of the response object.

common/mixins.py
nikomakela: See https://github.com/City-of-Helsinki/palvelutarjotin/pull/384#discussion_r2026770812

organisations/models.py
```diff
@@ -110,15 +113,69 @@ class Meta:
...
-    def delete(self, *args, **kwargs):
-        self.delete_related_p_event_contact_info()
-        return super().delete(*args, **kwargs)
+    def _delete_related_objects(self, using=None, keep_parents=False):
+        """
+        Deletes all related objects of the current instance.
+
+        This method replicates the logic of Django's internal delete mechanism
+        to delete related objects without deleting the instance itself.
+
```
karisal-anders, line 122: Linking to the source on which this function is based would bring more clarity/transparency:
```suggestion

        Based on Django's Model.delete method:
        https://github.com/django/django/blob/4.2.20/django/db/models/base.py#L1123-L1132

```

karisal-anders (approved):

terovirtanen:
**KULTUS-API branch is deployed to platta: https://kultus-pr384.api.dev.hel.ninja** :rocket::rocket::rocket:

PR #395, opened 2025-05-08
fix: event contact person should be publicly available

nikomakela:
PT-1898.

Fix an issue showing the event contact person information.

Add a `PublicPersonNode` on side with `PersonNode`. In `PublicPersonNode`, the access is not restricted only for staff. It's publicly readable, since it's reference to public event information.

image

### DIFF
--- a/occurrences/schema.py
+++ b/occurrences/schema.py
@@ -51,7 +51,7 @@ from occurrences.schema_services import (
 )
 from organisations.decorators import event_staff_member_required
 from organisations.models import Organisation
-from organisations.schema import PersonNodeInput
+from organisations.schema import PersonNodeInput, PublicPersonNode
 from palvelutarjotin.exceptions import (
     AlreadyJoinedEventError,
     ApiUsageError,
@@ -193,6 +193,7 @@ class PalvelutarjotinEventNode(DjangoObjectType):
         "ACCEPT-LANGUAGE header "
     )
     translations = graphene.List(PalvelutarjotinEventTranslationType)
+    contact_person = graphene.Field(PublicPersonNode)
 
     class Meta:
         model = PalvelutarjotinEvent
--- a/organisations/tests/test_api.py
+++ b/organisations/tests/test_api.py
@@ -4,6 +4,7 @@ import pytest
 from django.core import mail
 from django.utils import timezone
 from freezegun import freeze_time
+from graphene.test import Client
 from graphql_relay import to_global_id
 
 from common.tests.utils import assert_match_error_code, assert_permission_denied
@@ -13,7 +14,11 @@ from occurrences.factories import (
     PalvelutarjotinEventFactory,
 )
 from organisations.factories import OrganisationFactory, PersonFactory, UserFactory
-from organisations.models import Organisation, Person
+from organisations.models import (
+    Organisation,
+    Person,
+)
+from organisations.tests.schema import test_schema
 from palvelutarjotin.consts import (
     API_USAGE_ERROR,
     INVALID_EMAIL_FORMAT_ERROR,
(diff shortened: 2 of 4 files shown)


terovirtanen:
**KULTUS-API branch is deployed to platta: https://kultus-pr395.api.dev.hel.ninja** :rocket::rocket::rocket:

organisations/tests/test_api.py
```diff
@@ -909,3 +915,147 @@ def test_person_study_groups_unauthorized(
         variables={"id": to_global_id("PersonNode", person.id)},
     )
     assert executed["data"]["person"]["studygroupSet"]["edges"] == []
+
+
+@pytest.mark.django_db
+class PublicPersonNodeTests:
```
karisal-anders, line 921: Pytest doesn't run these tests, e.g. changing the class name to start with Test made the tests runnable:
```suggestion
class TestPublicPersonNode:
```

organisations/tests/test_api.py
```diff
@@ -909,3 +915,147 @@ def test_person_study_groups_unauthorized(
...
+            name="Private Person",
+            phone_number="333-333-3333",
+            email_address="private@example.com",
+            user=users["inactive"],
+        )
+        return {
+            "public1": public_person1,
+            "public2": public_person2,
+            "private": private_person,
+        }
```
karisal-anders, line 956: How about adding a person without linked user i.e. one of the filtered out cases in PublicPersonNode.get_queryset?
```suggestion
        private_person1 = Person.objects.create(
            name="Private Person One",
            phone_number="333-333-3333",
            email_address="private1@example.com",
            user=users["inactive"],
        )
        private_person2 = Person.objects.create(
            name="Private Person Two",
            phone_number="444-444-4444",
            email_address="private2@example.com",
            user=None,  # No linked user
        )
        return {
            "public1": public_person1,
            "public2": public_person2,
            "private1": private_person1,
            "private2": private_person2,
        }
```

And
```
        assert "Private Person" not in names
```
to
```
        assert "Private Person One" not in names
        assert "Private Person Two" not in names
```
later?

occurrences/schema.py
```diff
@@ -193,6 +193,7 @@ class PalvelutarjotinEventNode(DjangoObjectType):
         "ACCEPT-LANGUAGE header "
     )
     translations = graphene.List(PalvelutarjotinEventTranslationType)
+    contact_person = graphene.Field(PublicPersonNode)
```
karisal-anders, line 196: Although this is a one to one mapping i.e. PalvelutarjotinEventNode <-1:1-> PublicPersonNode, is there a query for fetching multiple PalvelutarjotinEventNodes and through them multiple PublicPersonNodes with a single query? If there is, then that's something that might be something that's not wanted because even though an event's contact person's name, email and phone number are public information, fetching them all with a single query may be unwanted (It may sound weird/unintuitive but that's the impression I got from an earlier discussion related exactly to this data's publicity).

If there is no such single query to fetch these en masse through nesting, then this is IMHO ok currently. But if there is any single query through which one can through node nesting fetch these, then it might not be ok, and I suggest we could have a talk about it as a team tomorrow in daily, or today still if people are present.

nikomakela: Yes, fetching events with their contact persons is possible, but it requires constructing a specific query ("a list of events with their contact persons"). From my perspective, there isn't a significant difference in effort or security between creating one such query or using two queries to achieve the same result by merging the data. The underlying contact information is publicly accessible in both scenarios. You could even easily open events in multiple browser windows to see details.

Therefore, I see the primary decision as: either we fix the bug to display the contact details on the event details page as intended, or we don't (which would then necessitate admin permissions to view them at all). With this PR, we are currently only showing these details on the individual event details page, not in any list view.

This PR directly addresses the bug requested by the product owner. While I'm not blocked from merging this immediately. We still have time to have conversations related to this, if we must, but I think then we need a product owner to make the decision.

karisal-anders: Invited the product owner to tomorrow's daily, so we could discuss this then

terovirtanen:
**KULTUS-API branch is deployed to platta: https://kultus-pr395.api.dev.hel.ninja** :rocket::rocket::rocket:

nikomakela:
Closed in favor of https://github.com/City-of-Helsinki/palvelutarjotin-ui/pull/411

PR #409, opened 2025-06-04
feat: remove persons with cronjob when retention period has exceeded

nikomakela:
PT-1932.

The `delete_retention_period_exceeding_contact_info` cronjob has deleted contact info from PalvelutarjotinEvent when the retention period (48h) has exceeded, but the persons have been deleted manually from the enrollee personal data admin view. Since it's been a manual task, it has always needed some one to take care of it. There is no reason why that job could not be done automatically with a cronjob.

The `delete_retention_period_exceeding_contact_info` cronjob is now extended with that feature, but there are also new flags that can be used to run only 1 of the tasks (with the management  command that the cronjob executes) and tehre is also a flag to make a dry drun, that does not acutally delete anything, but just reports how many instances it would delete if it would be ran.

### DIFF
--- a/occurrences/management/commands/delete_retention_period_exceeding_contact_info.py
+++ b/occurrences/management/commands/delete_retention_period_exceeding_contact_info.py
@@ -1,23 +1,144 @@
 from django.core.management.base import BaseCommand
 
 from occurrences.models import PalvelutarjotinEvent
+from organisations.models import EnrolleePersonalData
 
 
 class Command(BaseCommand):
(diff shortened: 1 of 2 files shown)


terovirtanen:
**KULTUS-API branch is deployed to platta: https://kultus-pr409.api.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
**KULTUS-API branch is deployed to platta: https://kultus-pr409.api.dev.hel.ninja** :rocket::rocket::rocket:

PR #422, opened 2025-08-22
PT-1946: Send notification only when person exists

mhieta:
Check also that the person object exists before sending the notification.

Fixes this issue: https://sentry.hel.fi/organizations/city-of-helsinki/issues/91881/?project=115

Refs: PT-1946


terovirtanen:
**KULTUS-API branch is deployed to platta: https://kultus-pr422.api.dev.hel.ninja** :rocket::rocket::rocket:

voneiden (changes requested):
I would prefer fixing it in the caller of this function ( https://github.com/City-of-Helsinki/palvelutarjotin/blob/55a5ef62ae41da117d904e18d907cbc617819d19/occurrences/models.py#L1041 ) - so add something like 

```python
if self.person is None:
    contact_people = []
else:
    contact_people = [self.person]
```

voneiden (approved):
Alrighty. Sonar doesn't seem to update the PR review but I think we can agree the report is a false positive so it's fine to merge.

